PR #22, opened 2017-03-19
C extensions (*.so) and build directory in gitignore list and Quickstart updated

nazrulworld:
When I run buildout to prepare guillotina development environment, I saw those `*.so` and build directory are generated. 
I think it would be good, if we put those in .gitignore

Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ include/
 pip-selfcheck.json
 .coveralls.yml
 docs/build
+*.so
+build/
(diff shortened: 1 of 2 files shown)


nazrulworld:
As we saw in Quickstart doc, there were some links and command (pcreate) from `plone.server` . We need to update according to `guillotina` for example `pcreate` is now `gcreate`

vangheem:
Thanks!

PR #26, opened 2017-03-22
add missing required param for sharing at quickstart docs

nazrulworld:
Required param `type` was missing for `@sharing` at quickstart doc, I added it with value `Allow`


vangheem:
Thanks!

PR #32, opened 2017-03-23
small development fixes to gitignore and Manifest.in

loechel:
while for guillotina, virtualenv is used, all such path should be excluded via gitignore, I have undated gitignore to be more detailed about excluded and included path

Also there was one Warring in Manifest.in on buildout, fixed it.


vangheem:
Thanks!

PR #44, opened 2017-04-06
WIP: Reuse transaction objects when doing `tm.begin()`

vangheem:


vangheem:
We sort of have issues around transactions...

https://github.com/plone/guillotina/commit/1728f18eaa816a93cfc15da47a76a3857037cdb9 fixed a lot of it but I'm not convinced.

We might need to look at how we're managing transactions and what the expectations are around it.

do we want to be able to manage multiple simultaneous transactions?

Right now we have a transaction stack that we push and pull off of for committing/aborting...

PR #52, opened 2017-04-14
Fix memory leak in get_current_request

vangheem:
Also some additional cleanup of unused code.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,13 @@
 1.0.0a11 (unreleased)
 ---------------------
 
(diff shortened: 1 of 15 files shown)


vangheem:
@bloodbare Would you please review this?

The fix here was removing `Py_INCREF(request);`.

Seems like we should need it because the request object already has references to it and we aren't actually creating a new reference here. We're just finding the object and if we assign it to another object, then the ref count would need to be increased.

We have no way to decrease the ref to it later so gc can clean it up either.

Let me know if I misunderstand anything.

References: https://docs.python.org/2/c-api/refcounting.html

PR #61, opened 2017-04-26
Provide robust conflict resolution on resources and annotations

vangheem:


vangheem:
@bloodbare would you be able and take a look to see what you think?

PR #65, opened 2017-05-03
Postgres simultaneously running statements issue

vangheem:
This is an exploration on potential issues with asyncio and trying to run simultaneous queries over a single connection.

Right now, it's just a group of failing tests and an attempted implementation trying to handle the error in the postgres storage module that shows it retrying doesn't work.

Tests to run....
- ./bin/pytest guillotina/tests/test_postgres.py -k test_using_gather_with_queries_before_prepare
- ./bin/pytest guillotina/tests/test_postgres.py -k test_using_gather_with_queries_after_prepare
- ./bin/pytest guillotina/tests/test_postgres.py -k test_exhausting_pool_size

Read the notes written in these tests about what error to expect. In the test_exhausting_pool_size case, you need to uncomment to see the behavior.

# However...

I wonder if this is not actually a problem. Is it a reasonable limitation to assume your simultaneous connections are limited to number the pool_size? The implementation still waits for connections to open up(except the @ws is a problem we need to fix with this PR having an implementation of fix for it).

Then, we'd just need to document that using things like `asyncio.gather` with db queries is unsafe. Also, `request._futures` would need to know this limitation as well. `request._futures` futures should already not be using the request's connection anyways because the transaction is finished before they are run.


vangheem:
This is fixed by using a lock on the transaction.

PR #90, opened 2017-05-26
Put restrictions on what valid ids for content are

vangheem:


guillotina/utils.py
```diff
@@ -275,6 +275,14 @@ def merge_dicts(d1: dict, d2: dict) -> dict:
     return result
 
 
+_valid_id_characters = string.digits + string.ascii_lowercase + '.-_'
+
+
+def valid_id(_id):
+    _id = _id.lower()
+    return _id == ''.join([l for l in _id if l in _valid_id_characters])
```
bloodbare: @vangheem could we improve the performance of this operation, joining a list is expensive, maybe a regex ?

PR #92, opened 2017-05-31
Adding queries to ask all content from one type and the total

bloodbare:


guillotina/db/storages/pg.py
```diff
@@ -546,3 +553,18 @@ def _db_transaction_factory(self, txn):
...
         return result
+
+    async def get_total_resources_of_type(self, txn, type_):
+        async with txn._lock:
+            smt = await txn._db_conn.prepare(NUM_RESOURCES_BY_TYPE)
+            result = await smt.fetchval(type_)
+        return result
+
+    # Massive treatment without security
+    async def _get_resources_of_type(self, txn, type_):
```
vangheem: I would recommend using something like the iterate_keys method. Trust me, on large data sets, a cursor will die on you.

PR #145, opened 2017-10-16
some fixes

thet:
... as i work with it.

- Fix Py_DECREF typo in optimizations.c. (c compiler complained while building with buildout)
- Let buildout depend on cookiecutter to be able to create project skeletons.


buildout.cfg
```diff
@@ -15,6 +15,7 @@ recipe = zc.recipe.egg
 eggs =
     aioconsole
     guillotina
+    cookiecutter
```
thet: not sure, if setup.py should depend on it because setup.py also installs the console_scripts...

buildout.cfg
```diff
@@ -62,6 +62,7 @@ eggs =
     pytest-cov
     pytest-watch
     guillotina [test]
+    cookiecutter
```
thet: not sure, if setup.py should depend on it because setup.py also installs the console_scripts...

vangheem:
Thanks @thet! I'll fix the tests.

vangheem:
Yah, I don't want it setup.py--it should be optional.

vangheem:
Closing, going to fix with another PR. optimizations is another issue that we're aware of.

PR #167, opened 2017-10-25
2.0/Master: various performance improvements and some API changes

vangheem:
- Make components more pythonic:
    - guillotina.component.getMultiAdapter -> get_multi_adapter
    - guillotina.component.getAdapter -> get_adapter
    - guillotina.component.getUtility -> get_utility
    - guillotina.component.queryUtility -> query_utility
    - guillotina.component.getUtilitiesFor -> get_utilities_for
    - guillotina.component.getAllUtilitiesRegisteredFor -> get_all_utilities_registered_for
    - guillotina.component.getGlobalSiteManager -> get_global_components
    - guillotina.component.provideUtility -> provide_utility
  [vangheem]

- Allow passing arguments to component lookup factories with `args=[]` and `kwargs={}` params
  [vangheem]

- Optimize serialization framework to use less and faster adapter lookups
  [vangheem]

- Added `guillotina.profile.profilable` decorator to be able to tell line_profiler
  what functions to profile.

- Integrate line_profiler with `--line-profiler`, `--line-profiler-matcher="*foobar*"`
  and `--line-profiler-output`.
  [vangheem]

- Pre-render resolve_dotted_name on authentication plugins
  [vangheem]

- Implement aiotask_context for request object lookup
  [vangheem]

- Add `@configure.value_deserializer` and `@configure.value_serializer`
  configurators
  [vangheem]

- Remove transform framework
  [vangheem]

- Remove RichText since it was not used and didn't make much sense
  [vangheem]


guillotina/behaviors/properties.py
```diff
@@ -10,19 +10,11 @@ def __get__(self, inst, klass):
         if inst is None:
             return self
 
-        if hasattr(inst.context, self.__name__):
```
bloodbare: Why we are removing this check ?

guillotina/component/hooks.py
```diff
@@ -14,13 +14,13 @@
...
-import contextlib
-import threading
-
-from guillotina.component.globalregistry import getGlobalSiteManager
+from guillotina.component.globalregistry import get_global_components
 from guillotina.component.interfaces import ComponentLookupError
 from guillotina.component.interfaces import IComponentLookup
 
+import contextlib
+import threading
```
bloodbare: Can we remove the SetSite and the threading import ?

guillotina/db/transaction.py
```diff
@@ -76,10 +78,10 @@ def __init__(self, manager, request=None, loop=None):
         # we *not* follow naming standards of using "_request" here so
         # get_current_request can magically find us here...
         self.request = request
-        self._strategy = getMultiAdapter(
+        self._strategy = get_multi_adapter(
```
bloodbare: Should we move this two get_multi_adapter to single adapters?

guillotina/traversal.py
```diff
@@ -350,7 +354,7 @@ def set_root(self, root):
 
         await self.apply_authorization(request)
 
-        translator = queryMultiAdapter(
+        translator = query_multi_adapter(
```
bloodbare: Do we have profiling information about the three multiadapters on the real_resolve ?

guillotina/behaviors/properties.py
vangheem: Because we always want to set the value even if it didn't previously have a value on the object.

guillotina/component/hooks.py
vangheem: I thought I'd leave it because I was worried about what happens with executors...

guillotina/traversal.py
vangheem: Yes, we have some. I would be in favor of providing a different mechanism to hook into processing requests instead of doing translations on every request.

vangheem: I would be in favor of removing the translation framework since it's undocumented and unused.

We can do something in an addon.

guillotina/behaviors/properties.py
bloodbare: Isn't it that in case that we have the AttributeError we were displaying the default value ? Setting it will make things harder to change the default value of a field.

guillotina/component/hooks.py
bloodbare: I can't understand why its involved with executors.

guillotina/traversal.py
bloodbare: Its not only the translator aproach, also the view and renderer multiadapter. We should have an adapter so it can be implemented on addons without the need to modify real_traversal.

guillotina/behaviors/properties.py
vangheem: We are using the default value here though `self.default`

guillotina/component/hooks.py
vangheem: What is wrong with thread locals?

PR #197, opened 2017-12-17
better test coverage

vangheem:


guillotina/utils.py
```diff
@@ -307,32 +279,6 @@ def get_current_request() -> IRequest:
     raise RequestNotFound(RequestNotFound.__doc__)
 
 
-def apply_cors(request: IRequest) -> dict:
```
bloodbare: Where is the new Apply Cors ?

vangheem: The CORs renderer configurable. See:

- https://github.com/plone/guillotina/blob/master/guillotina/cors.py
- https://github.com/plone/guillotina/blob/master/guillotina/factory/app.py#L75
- https://github.com/plone/guillotina/blob/master/guillotina/traversal.py#L291

PR #198, opened 2017-12-20
Implement transaction strategies to delay starting db transaction to just before commit phase

vangheem:
This addresses the recommendations of this document yet: https://www.cockroachlabs.com/docs/stable/transactions.html#client-side-intervention


vangheem:
@bloodbare this PR makes everything serializable isolation level since it seems like we should stay away from snapshot because of how unpredictable it is.

PR #205, opened 2018-01-09
Added "Class-based services" example

masipcat:


vangheem:
Oops! Thanks for fixing the docs!

PR #210, opened 2018-01-15
Add `Item.async_multi_get` method

vangheem:


guillotina/db/transaction.py
```diff
@@ -407,8 +452,8 @@ def tpc_cleanup(self):
     async def items(self, container):
         # XXX not using cursor because we can't cache with cursor results...
         keys = await self.keys(container._p_oid)
```
bloodbare: Would not be better to have an async generator for the keys and get the batches with the keys?

guillotina/db/transaction.py
```diff
@@ -384,16 +384,61 @@ def tpc_cleanup(self):
...
+        - async for iterate items
+        - store retrieved values in storage
+        '''
+        lookup_group = []  # backlog of object that need to be looked up
+        for key in keys:
+            item = await self._cache.get(container=parent, id=key)
+            if item is None:
+                self._cache._misses += 1
+                lookup_group.append(key)
+                if len(lookup_group) > 15:  # limit batch size
```
bloodbare: Why this limit, can it be configured ?

vangheem: async generators in postgresql queries are not ideal. They require cursors. Cursors require an active transaction and we often don't have that.

guillotina/db/transaction.py
vangheem: async generators in postgresql queries are not ideal. They require cursors. Cursors require an active transaction and we often don't have that.

guillotina/db/transaction.py
vangheem: for example, getting all the objects in a large container.

PR #218, opened 2018-01-22
Config file path

cdevienne:


cdevienne:
Hi @vangheem, it seems this feature disappeared in 2.5.x, is it on purpose?

cdevienne:
It's actually in 3.0.0a1 that __file__ is absent of app_settings.

vangheem:
I see it here: https://github.com/plone/guillotina/blob/master/guillotina/commands/__init__.py#L68

vangheem:
@cdevienne gotcha. Nevermind.

cdevienne:
After an upgrade the entry was not present when starting a service.
I reverted to 2.5 (and master) anyway, so I am good for now.

PR #225, opened 2018-02-06
Deserializer for IDate

cdevienne:


guillotina/json/deserialize_value.py
```diff
@@ -119,3 +121,10 @@ def datetime_converter(field, value, context):
     if not isinstance(value, str):
         raise ValueDeserializationError(field, value, 'Not a string')
     return parse(value)
+
+
+@configure.value_deserializer(IDate)
+def date_converter(field, value, context):
+    if not isinstance(value, str):
+        raise ValueDeserializationError(field, value, 'Not a string')
+    return datetime.datetime.strptime(value, '%Y-%m-%d').date()
```
vangheem: Think we should do `parse(value).date()` instead?

cdevienne: We could, but it would accept not only date but datetime, which is not a good thing imho

PR #236, opened 2018-03-14
Implement patch and bucket list fields

vangheem:


guillotina/fields/annotation.py
```diff
@@ -0,0 +1,203 @@
...
+        annotation['items'].append(value)
+        annotation._p_register()
+
+    async def extend(self, context, value):
+        for item in value:
+            await self.append(context, item)
+
+    def __len__(self):
+        total = 0
+        for metadata in self.annotations_metadata.values():
```
bloodbare: If the amount of buckets is really large we may have a lot of loads to get the total?

vangheem: This is a small metadata object. If you have 1 million buckets, it might be slow but if you're using this to store billions of items(bucket size * # of buckets), you're doing something wrong...

vangheem: Can you approve?

PR #241, opened 2018-03-19
Remove dependency on blobs/objects

bloodbare:


vangheem:
What is the problem here? `ALTER TABLE blobs DROP CONSTRAINT IF EXISTS blobs_zoid_fkey;` has an "IF EXISTS" statement so it should be harmless.

bloodbare:
Blobs table does not exist if blobs sql is not run

vangheem:
Well, this is going to cause other tests to fail. Can you change it to an `ALTER TABLE blobs IF EXISTS`?

PR #271, opened 2018-05-18
Indexing updates

vangheem:


guillotina/patch.py
```diff
@@ -16,7 +16,8 @@
 async def asubscribers(self, objects, provided):
     subscriptions = self.subscriptions(map(providedBy, objects), provided)
     results = []
-    for subscription in subscriptions:
+    for subscription in sorted(subscriptions,
+                               key=lambda sub: getattr(sub, 'priority', 100)):
```
bloodbare: Can we find a way to avoid sorting the events? could we store them sorted at register time?

PR #280, opened 2018-05-30
Make sure all responses from the API are json

vangheem:


vangheem:
@bloodbare It doesn't handle raised exceptions.

Also, I think we should consider removing this "feature"...

PR #281, opened 2018-06-04
Response/Rendering fixes

vangheem:
- `guillotina.browser.Response` moved to `guillotina.response.Response`
- move `guillotina.browser.ErrorResponse` to `guillotina.response.ErrorResponse`
- `guillotina.browser.UnauthorizedResponse` removed
- `guillotina.response.Response` no longer supports wrapping aiohttp responses
- `guillotina.response.Response` can now be raised as an exception
- return or raised aiohttp responses now bypass guillotina renderer framework
- raising any Response as an exception aborts current transaction
- remove `IFrameFormatsJson`
- remove `IRenderFormats`, `IRendered` is now a named adapter lookup
- remove `app_settings.renderers` setting. Use the lookups
- remove `IDownloadView`
- remove `TraversableDownloadService`
- remove `IForbiddenAttribute`
- remove `ISerializableException`
- remove `IForbidden`


guillotina/contentnegotiation.py
```diff
@@ -1,844 +1,34 @@
...
+    ca-ES,ca;q=0.9,en;q=0.8,es;q=0.7
+    '''
+    result = []
+    for ct in value.lower().split(','):
+        weight = 1.0
+        ct_name = ct.split(';')[0]
+        for param in ct.split(';')[1:]:
+            if param.startswith('q='):
+                try:
+                    weight = float(param[len('q='):])
```
bloodbare: Can we replace to 2: ?

guillotina/contentnegotiation.py
```diff
@@ -1,844 +1,34 @@
...
+                except ValueError:
+                    pass
+        result.append((weight, ct_name))
+    return [i[1] for i in sorted(result, reverse=True)]
```
bloodbare: maybe we could use the _sort_by_q deleted function so we get the real order?

CHANGELOG.rst
```diff
@@ -1,7 +1,21 @@
...
-- Nothing changed yet.
+4.0.0 (unreleased)
+------------------
+
+- `guillotina.browser.Response` moved to `guillotina.response.Response`
+- move `guillotina.browser.ErrorResponse` to `guillotina.response.ErrorResponse`
+- `guillotina.browser.UnauthorizedResponse` removed
+- `guillotina.response.Response` no longer supports wrapping aiohttp responses
+- `guillotina.response.Response` can now be raised as an exception
+- return or raised aiohttp responses now bypass guillotina renderer framework
```
ableeb: typo on return should be 'returned'

guillotina/contentnegotiation.py
```diff
@@ -1,844 +1,34 @@
...
-        return self.cn.negotiate(**kwargs)
-
-
-def content_type_negotiation(request, resource, view):
+def get_ordered_content_type_value_list(value):
+    '''
+    ca-ES,ca;q=0.9,en;q=0.8,es;q=0.7
+    '''
+    result = []
+    for ct in value.lower().split(','):
```
bloodbare: We should strip also

guillotina/contentnegotiation.py
```diff
@@ -1,844 +1,34 @@
...
+        weight = 1.0
+        ct_name = ct.split(';')[0]
+        for param in ct.split(';')[1:]:
```
bloodbare: There is the use case were accept can be:     Accept: text/*, text/plain, text/plain;format=flowed, */*
[https://tools.ietf.org/html/rfc7231#section-5.3.2](https://tools.ietf.org/html/rfc7231#section-5.3.2)

guillotina/contentnegotiation.py
```diff
@@ -1,844 +1,34 @@
...
-        # if download view, we want to render raw immediately
-        # or if no or */* accept header provided
-        return IRendererFormatRaw
-
-    np = get_utility(IContentNegotiation, 'content_type')
-    ap = np.negotiate(accept=accept)
-    # We need to check for the accept
-    if str(ap.content_type) in app_settings['renderers']:
-        return app_settings['renderers'][str(ap.content_type)]
+        return get_ordered_content_type_value_list(request.headers['ACCEPT'])
```
bloodbare: We are losing the difference between type and subtype. Maybe we need two different parsers (like the code we are removing)

guillotina/contentnegotiation.py
vangheem: awww, I like this more. Compiler takes care of it and makes `'q='` a constant to 2 anyways.

It's a style thing to be explicit about what we're doing.

guillotina/contentnegotiation.py
vangheem: I'm not familiar with what you're talking about here. Can you clarify?

bloodbare: If u check the function from the old code u will see the logic on the priority to define which is prior to which one

vangheem: ah yes, I see. I'll fix

vangheem: the old code is so horrible

PR #293, opened 2018-07-13
Allow patching registry with new Interface fields

jordic:
When evolving an interface on the registry, allow to patch with new
fields on the interface


vangheem:
flake8 error: `guillotina/tests/test_api.py:194:1: E303 too many blank lines (3)`

PR #311, opened 2018-07-23
ship Executioner with Guillotina

ebrehault:
Note: If you think we must ship it as an external package, let me know.


ebrehault:
> If there any way to ship where the svg files are combined?

I am not sure I understand, do you mean here https://github.com/plone/guillotina/tree/master/static/brand ?

> I'm going to tweak this so the files are in the python module and can be used from other contexts without needing to know the full path.

Ok, does it implies the publication path is not necessarily `/+admin/`? If so, we'll need to change the Angular build.

vangheem:
I'm sorry "if" should have been "is".

It will still be served from same URL so we're good!

ebrehault:
So I just have to move `/static/executioner` to `/static/brand/executioner`, right?

vangheem:
I fixed with: https://github.com/plone/guillotina/pull/311/commits/2bb5edf7cb3e505ea67a69b56f541d77acb43e8f

PR #313, opened 2018-07-23
remove router from aio app constructor.

jordic:
As I understood on the original aiohttp issue, seems that this could solve the problem. 
But it's not so clear for me, as seems, like they delegates the router creation to the wrapped App.
What's your thought on it? @vangheem

### DIFF
--- a/guillotina/factory/app.py
+++ b/guillotina/factory/app.py
@@ -65,6 +65,10 @@ def load_application(module, root, settings):
 
 
 class GuillotinaAIOHTTPApplication(web.Application):
+    def __init__(self, router, middlewares, **kwargs):
+        super().__init__(middlewares=middlewares, **kwargs)
+        self._router = router
+
     async def _handle(self, request, retries=0):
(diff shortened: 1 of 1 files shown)


vangheem:
This will prevent the warning; however, I don't think it will fix the problem. They are removing the router infrastructure.

jordic:
That's something I also understood... but anyway seems wired to have a webframework without router infraestructure... :/

bloodbare:
I agree that the aproach would need to be router /* to guillotina router, still no need to merge until aiohttp 4

bloodbare:
Just to keep track a reference to aiohttp issue: https://github.com/aio-libs/aiohttp/issues/2766

vangheem:
@bloodbare what do you think about merging this still so our logs get cleaned up on test runs?

@jordic thoughts?

jordic:
will ask it on the aio gitter.

vangheem:
Closing ticket for now until we decide on a solution.

Issue #308, opened 2018-07-22
DB api improvements proposals


cdevienne:
This issue gather the ideas we talked about during the Costa-Brava sprint.

### Simple transactions

Simplify the transaction opening api. It should always be a one line operation.

```python
txn = await tm.begin()
await txn.commit()
```

If the tm is associated with a request, the it should be passed to the
transaction, which removes any need to explicitely pass the request.

### Request default transaction

It the transaction is to become the default transaction of a request, the
eventual container of the request should be moved to the new transaction,
or refetched, and the transaction be set as the request.\_txn:

```python
txn = await request.begin()
await txn.commit()
```

Also, if the transaction is the default request transaction, commiting it
should call `execute_futures()` automatically.

### Context manager

A transaction manager for handling transactions could be provided (it
might already be the case):

```python
async with transaction(tm):
    pass
```

## Flush

In some cases it is needed to apply all changes to the database without closing
the transactions. There is currently no way to do that with the db api.

The proposal is to add a `flush()` function to Transaction, that does what
tpc_commit does (except for commit itself of course). This function would be
called automatically on commit.

```python
async txn.flush()
```

### Problems

When a object is added, or its path changed, the navigation api (async_get, 
navigate_to) will not 'see' it until the changes are flushed to the database.

Implementing flush somehow fix the problem but at a double cost:

- the user must remember to flush before navigating
- a same object my be loaded/modified/flushed many times in a single transaction.

Another problem with the current implementation is that a same object can be
loaded several times as different instances, and loose all consistency between
those instances.

### Proposal

The idea is to add a path-based index of all the objects that were loaded.

This index would be used by the content api (async_get etc.), and achieve
complete consistency of the objects loaded in a transaction.

The index would use weakref so that the memory footprint is as low as possible
(we don't want to keep an object loaded when its not used anywhere).

cdevienne:
After attempting to implement flush(), and some discussions with Nathan and Ramon, it appeared that adding a flush() function is not a good idea.

The reason is that guillotina has to try very hard to keep transactions as short as possible, to maximize the cockroach efficiency. Having a flush() will create more problems than it could solve.

-> Add flush() abandoned!

vangheem:
We added a content api.

I think the main take away left from this is to provide a well-documented context manager for transactions. We have a context manager but I think we can simplify it more.

Closing for this https://github.com/plone/guillotina/issues/437

PR #315, opened 2018-07-23
Add Navigator utility

cdevienne:
It implements a path-based identity map, as suggested in #308

### DIFF
--- a/guillotina/utils/navigator.py
+++ b/guillotina/utils/navigator.py
@@ -1,0 +1,47 @@
+import posixpath
+import weakref
+
+from .content import get_content_path
+from guillotina.interfaces import IResource
+
+
+class Navigator:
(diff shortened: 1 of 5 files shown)


guillotina/utils/navigator.py
```diff
@@ -0,0 +1,84 @@
...
+        self.txn.delete(obj)
+        del self.index[path]
+
+    def _load_obj(self, obj: IResource):
+        """insert an object in the index"""
+        if obj.__parent__ and get_content_path(
+                obj.__parent__) not in self.index:
+            self._load_obj(obj.__parent__)
+        self.index[get_content_path(obj)] = obj
+        print(list(self.index.keys()))
```
bloodbare: Should not be a print on the code

PR #316, opened 2018-07-24
Adding tests on cookiecutter

bloodbare:


vangheem:
Looks awesome but we have build failures!

PR #321, opened 2018-08-05
Compatibility

bloodbare:


guillotina/files/field.py
```diff
@@ -145,11 +147,18 @@ def extension(self, val):
...
-        val = await file_manager.save_file(
-            partial(_generator, value), content_type=value['content_type'],
-            size=len(value['data']))
-        return val
+
+    # There is not file and expecting a dict
+    # 'data', 'encoding', 'content-type', 'filename'
+    request = get_current_request()
+    file_manager = get_multi_adapter((context, request, field), IFileManager)
+    content_type = value['content_type'] if 'content_type' in value else value['content-type']
```
vangheem: This looks wrong.

Can we just use `value.get('content-type')`?

bloodbare: The problem is that we used to have content_type as the key and plone rest api uses content-type

PR #323, opened 2018-08-21
Adding support to hook before modification of a field

bloodbare:


guillotina/events.py
```diff
@@ -149,3 +150,11 @@ class NewUserAdded(object):
 
     def __init__(self, user):
         self.user = user
+
+
+@implementer(IFieldPreModified)
+class FieldPreModifiedEvent(object):
```
vangheem: Can we change the name to "BeforeFieldModifiedEvent" so it matches the scheme or our other events?

bloodbare:
@vangheem  can you review again?

PR #327, opened 2018-08-28
PatchField for ints to implement counter

lferran:


vangheem:
Does this work on complex objects? Like ones with the schema.Object type?

Can you look over your implementation and see why the others use `self.get_value` and let me know? Or add a test which is testing against a complex object type like:

```
class IFoo(Interface):
    int_val = schema.Int()

class IBar(Interface):
    foo = schema.Object(IFoo)
```

lferran:
For now, this was just supposed to work for Int attributes. Why would we want to support complex objects?

vangheem:
Just to be consistent and so users aren't confused by it I guess. @bloodbare thoughts?

Could you also please add documentation for the new feature here: https://guillotina.readthedocs.io/en/latest/developer/fields.html

Thank you!

lferran:
Another question: do you think it's a good idea to allow `inc`, `dec` and `reset` operations to be issued without the `value` on the payload? so that, e.g:
 - `inc` increments by 1 unit by default
 - `dec` decrements by 1 unit by default
 - `reset` resets the value to field's `default` (or `None` if not set) 

or would it better to stick to forcing `value` to be present? (from an API point of view)

vangheem:
Yes, I think it'd be great to have it without a value!

bloodbare:
@lferran having orthogonal API is one of the musts on it, if we start to have use cases that are not covered will be complex for people to understand, so it should support complex objects.

About the inc, dec and reset its a great aproach,

lferran:
@bloodbare :+1: 

so what you guys mean is that, when having:

```
class IFoo(Interface):
    int_val = PatchField(schema.Int())
    other_field = TextLine()

class IBar(Interface):
    foo = schema.Object(IFoo)
    blah = TextLine()
```

I should be able to patch it like this:
```
{
'blah': 'world',
'foo': {
  'other_field': 'hello',
  'int_val': {
    'op': 'inc', 'value': 2
  }
}
```

Is that it?

vangheem:
Take a look at tests related to https://github.com/plone/guillotina/blob/master/guillotina/tests/test_serialize.py#L166 for example.

guillotina/fields/patch.py
```diff
@@ -41,10 +42,7 @@ def field_converter(field, value, context):
         if operation is None:
             raise ValueDeserializationError(
                 field, value, f'"{operation_name}" not a valid operation')
-        if 'value' not in value:
-            raise ValueDeserializationError(
-                field, value, f'Mising value')
-        value = operation(context, value['value'])
+        value = operation(context, value.get('value'))
```
vangheem: I'm a bit concerned about possible side-affects with defaulting to `None` here where maybe it should throw an error.

vangheem:
Merging. Nice job!

PR #333, opened 2018-08-30
Allow PUT in CORS policy

ebrehault:


config-dummy.yaml
```diff
@@ -5,10 +5,19 @@ databases:
   df:
     storage: DUMMY_FILE
     filename: dummy_file.db
-host: 0.0.0.0
+host: 127.0.0.1
```
vangheem: Can we remove the extra lines here?

PR #340, opened 2018-09-13
Non-defined fields in schema are stored (in some scenarios)

masipcat:
Test to reproduce the problem

### DIFF
--- a/guillotina/fields/patch.py
+++ b/guillotina/fields/patch.py
@@ -43,6 +43,8 @@ def field_converter(field, value, context):
             raise ValueDeserializationError(
                 field, value, f'"{operation_name}" not a valid operation')
         value = operation(context, value.get('value'))
+    elif isinstance(value, (dict, list)):
+        value = get_adapter(field.field, IJSONToValue, args=[value, context])
     return value
 
 
--- a/guillotina/tests/test_dynamic_schema.py
+++ b/guillotina/tests/test_dynamic_schema.py
@@ -366,7 +366,7 @@ async def test_add_dynamic_fields(container_requester):
                     "fields": {
                         "foobar": {
                             "title": "Hello field",
-                            "type": "int"
+                            "type": "integer"
                         }
                     }
                 }
@@ -470,7 +470,7 @@ async def test_add_dynamic_fields_invalid_type(container_requester):
                     "fields": {
                         "foobar": {
                             "title": "Hello field",
-                            "type": "int"
+                            "type": "integer"
                         }
                     }
                 }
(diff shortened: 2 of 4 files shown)


masipcat:
@bloodbare Yes, I'll send it today later ;)

guillotina/fields/patch.py
```diff
@@ -43,6 +43,8 @@ def field_converter(field, value, context):
             raise ValueDeserializationError(
                 field, value, f'"{operation_name}" not a valid operation')
         value = operation(context, value.get('value'))
+    else:
```
vangheem: Can we change this to maybe `elif isinstance(value, (dict, list)):`?

masipcat: sure!

masipcat:
@vangheem Can you review this comment https://github.com/plone/guillotina/pull/340#discussion_r217710688 ?

guillotina/fields/patch.py
vangheem: Why int?

masipcat: @vangheem ups, forget this :P

vangheem:
@masipcat we have test failures now :( https://travis-ci.org/plone/guillotina/jobs/428620976

Can you see?

guillotina/tests/test_dynamic_schema.py
```diff
@@ -491,7 +491,12 @@ def get_foo(inst):
                     }
                 }
             }))
-        assert status == 412
+        assert status == 204
```
masipcat: @vangheem  I changed this because I think that this call is correct... Can you confirm?

vangheem: hmmm, I don't think so. This is a test to specifically make sure it gives a configuration error if you provide the wrong type for a value.

Was this test always wrong maybe?

`foobar` is an int though and it's trying to patch it as a string.

masipcat: This test was passing in the past because the field type was `int` instead of `integer` and this was producing a 412 for any kind of value. If tried patching with a `5` (without quotes) and is also returning a 412.

So, after fixing the PatchField thing, guillotina started complaining that `int` is not a valid `type`.

Now, when doing the PATCH with a quoted integer the `IJSONToValue` transforms the string to integer without complaining https://github.com/plone/guillotina/blob/master/guillotina/fields/dynamic.py#L63 . I guess this is happening in other places. Is this behavior correct?

Also, I updated the tests to verify that a string like `"not an integer"` fails with 412

vangheem: cool, looks good then.

PR #342, opened 2018-09-13
be able to override configuration with command line options

vangheem:


guillotina/commands/__init__.py
```diff
@@ -75,6 +83,18 @@ def get_settings(configuration):
                            f'Using default settings.')
         MISSING_SETTINGS['logged'] = True
         settings = MISSING_SETTINGS.copy()
+
+    for override in overrides or []:
+        if '=' not in override:
+            raise Exception(f'Invalid configuration {override}')
```
lucid-0: Can we use f string in python 2?

PR #344, opened 2018-09-19
Static Behavior on serialize object

bloodbare:


guillotina/json/serialize_content.py
```diff
@@ -56,20 +56,24 @@ def __init__(self, context, request):
...
+            behaviors.append(behavior_schema)
+
         result = {
             '@id': IAbsoluteURL(self.context, self.request)(),
             '@type': self.context.type_name,
             '@name': self.context.__name__,
             '@uid': self.context.uuid,
             'parent': parent_summary,
             'is_folderish': IFolder.providedBy(self.context),
+            '__static_behaviors__': behaviors,
```
vangheem: Can we use a `@` value here?

Sorry, I don't like adding non-editable attributes without the `@`.

guillotina/json/serialize_content.py
```diff
@@ -56,20 +56,24 @@ def __init__(self, context, request):
...
+        behaviors = []
+        for behavior_schema in factory.behaviors or ():
+            behaviors.append(behavior_schema)
+
         result = {
             '@id': IAbsoluteURL(self.context, self.request)(),
             '@type': self.context.type_name,
             '@name': self.context.__name__,
             '@uid': self.context.uuid,
+            '@sb': behaviors,
```
vangheem: sorry, can we just call it "@static_behaviors"?

bloodbare: this information is going to be on all responses of GET so I was willing to keep it short

vangheem: eh, I don't know man. I think we need the name to be descriptive.

Up to you now. I'll approve but you're free to make final decision.

PR #346, opened 2018-09-19
Be able to register a behavior with only an interface

vangheem:


guillotina/json/serialize_content.py
```diff
@@ -67,11 +67,11 @@ def __init__(self, context, request):
             '@name': self.context.__name__,
             '@uid': self.context.uuid,
             '@static_behaviors': behaviors,
-            'parent': parent_summary,
-            'is_folderish': IFolder.providedBy(self.context),
+            'parent': parent_summary,  # should be @parent
+            'is_folderish': IFolder.providedBy(self.context),  # eek, should be @folderish?
```
bloodbare: https://plonerestapi.readthedocs.io/en/latest/content.html parent and is_folderish are not '@'

bloodbare: @timo ^

PR #348, opened 2018-09-19
Move static guillotina assets into python package

vangheem:


bloodbare:
Why we need moving assets from static to jsapps?

vangheem:
oops, I'll fix

PR #351, opened 2018-09-20
add @resolveuid endpoint

vangheem:


guillotina/api/content.py
```diff
@@ -781,13 +785,9 @@ def getRequestMethod(self):  # noqa
     if request.query.get('omit'):
         omit = request.query.get('omit').split(',')
 
-    security = IInteraction(request)
-
     results = []
     for key in await txn.get_page_of_keys(context._p_oid, page=page, page_size=page_size):
         ob = await context.async_get(key)
-        if not security.check_permission('guillotina.ViewContent', ob):
```
bloodbare: Can we provide another endpoint for this use case ?

PR #359, opened 2018-10-01
Provide support for jupyter notebooks

vangheem:


guillotina/api/content.py
```diff
@@ -178,11 +178,18 @@ class DefaultPOST(Service):
 
         user = get_authenticated_user_id(self.request)
 
+        options = {
+            'id': new_id,
+            'creators': (user,),
+            'contributors': (user,)
+        }
+        if 'uid' in data:
```
bloodbare: Why we need to be able to define uid on POST creation?

vangheem: In running migrations from plone, you need to match UIDs so linking works properly

PR #364, opened 2018-10-07
String is the serialize schema value for choice

bloodbare:


bloodbare:
@ebrehault this will fix the problem on PUT were Language should be a string not a list

PR #379, opened 2018-11-10
Improved index page of the docs

hirokiky:
I started guillotina in these days, and I thought the index page of the docs can be improved to find information.

* Fixed broken link
* We need to read the "training" before the "Narrative guide", but the link for "training" is too small

I fixed order of each information, like from shallow to deeper one.


hirokiky:
I rebased this branch and fixed the change log.

PR #381, opened 2018-11-10
Proposal to not load async utils by default and utils as dicts

bloodbare:


bloodbare:
1. Now you can’t define which utilities are loaded or reload them
2. Configuration is the same complexity as before
3. You can load multiple times utilities with multiple settings

bloodbare:
1. Now you can not define that an utility does not start if its loaded on the package.

2. With the new aproach u dont need to configure anything also.

3. There is two registrations right now, all async utils and the settings[‘utilities’], is confusing because they are loaded multiple times

PR #390, opened 2018-12-07
added and emit IBeforeRenderViewEvent

lferran:
needed in https://github.com/guillotinaweb/guillotina_ratelimit/pull/1

### DIFF
--- a/guillotina/events.py
+++ b/guillotina/events.py
@@ -7,6 +7,7 @@ from guillotina.interfaces import IBeforeFieldModifiedEvent
 from guillotina.interfaces import IBeforeObjectAddedEvent
 from guillotina.interfaces import IBeforeObjectMovedEvent
 from guillotina.interfaces import IBeforeObjectRemovedEvent
+from guillotina.interfaces import IBeforeRenderViewEvent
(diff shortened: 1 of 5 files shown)


vangheem:
Can we rename it to `IBeforeRenderView` and move it to here: https://github.com/plone/guillotina/blob/master/guillotina/traversal.py#L253

\cc @bloodbare

guillotina/events.py
```diff
@@ -72,6 +73,11 @@ class ObjectMovedEvent(ObjectLocationEvent):
     """An object has been moved"""
 
 
+@implementer(IBeforeRenderView)
+class BeforeRenderView(ObjectEvent):
```
vangheem: Can we rename to `BeforeRenderViewEvent`.

Do not inherit from `ObjectEvent`, it's not an event on an object!

Need to implement `def __init__(self, request)` method

lferran: oh ok!

lferran: can I pass the view object aswell? `def __init__(self, request, view)`
I'd like to use it to get the view's route name, but not sure if that already comes in the request

vangheem: yup!

PR #391, opened 2018-12-14
WIP: prefix dynamic storage

vangheem:


vangheem:
@bloodbare and I talked about the implementation and there are some concerns about complexity and performance.

- parameterize table names
- move prefix implementation to new package

vangheem:
fyi, pg does not allow table names to be parameterized...

So we're back to some form of sql string manipulation.

vangheem:
@bloodbare What do you think about an approach like this? https://github.com/plone/guillotina/commit/79fcdfe87978b040ba127c9bcf94753fd24407c3

bloodbare:
@vangheem I like the second aproach

Issue #393, opened 2018-12-21
Add key-value to schema.Dict fails silently


masipcat:
Hi! Here is my weekly issue/pr :P

I'm opening this issue because I found something weird with fields of type dict with default values (maybe list are affected as well)

## Example

I have a custom content-type like this:

```python
class ICustomContentType(IItem):
    images = schema.Dict(
        key_type=schema.TextLine(),
        value_type=schema.TextLine(),
        required=False,
        default={}
    )
```

And I implemented an endpoint that adds a key-value pair to field `images`.

```python
@configure.service(
    method="PATCH",
    name="@update-dict",
    permission="guillotina.AddContent"
)
async def update_default_dict(context, request):
    context.images['new-key'] = 'value'
    request._txn.register(context)
    assert len(context.images) == 0  # wat?
```

After this endpoint retrieve the object and the field `images` is `{}`.

**BUT** if I replace the previous code with `context.images = {'new-key': 'value'}` then everything works fine.

Is this a bug or I'm not using default value of field properly?

## Reproduce

I implemented a test that reproduces the problem: https://github.com/masipcat/guillotina/commit/9be9d11a073f89bb526a7ee63aedb2f8aaddbe15

Thank you!

vangheem:
Isn't that because using default of `{}` means that's like a global. Can you change that to `defaultFactory=dict` and see if that works?

vangheem:
functionally, sure; however, if you think about just regular python, providing a dict as a default param is dangerous.

For example, you should never do this:

```
def foobar(opt={}):
  opt['foobar'] = 5  # this value will stay in the default param forever
```

You should do this instead:

```
def foobar(opt=None):
  if opt is None:
    opt = {}
```

That was my point.

masipcat:
@vangheem yeah, I know, learned this the hard way in the past... hehe.
Also, I thought that guillotina was creating a copy of the `default` every time `__getattr__` is called to avoid sharing the same reference between objects.

vangheem:
Ahh, you're right. I forgot.

masipcat:
Hey, I think I found the problem

```python
160  	async def update_default_dict(context, request):
161  	    context.images['new-key'] = 'value'
162  	    import pdb; pdb.set_trace()
163  ->	    request._txn.register(context)
164  	    assert len(context.images) == 0  # wat?
(Pdb) id(context.images)
140677027687088
(Pdb) id(context.images)
140677020557888
(Pdb) id(context.images)
[...]

PR #395, opened 2018-12-24
Fix Resource.__getattr__() for empty fields with default values

masipcat:
Fixes https://github.com/plone/guillotina/issues/393

### DIFF
--- a/guillotina/tests/test_content.py
+++ b/guillotina/tests/test_content.py
@@ -9,11 +9,32 @@ from guillotina.content import load_cached_schema
 from guillotina.exceptions import NoPermissionToAdd
 from guillotina.exceptions import NotAllowedContentType
 from guillotina.interfaces import IApplication
+from guillotina.interfaces import IItem
 from guillotina.interfaces.types import IConstrainTypes
+from guillotina.schema import Dict
+from guillotina.schema import TextLine
 from guillotina.tests import utils
 
 import pickle
 import pytest
+
+
+class ICustomContentType(IItem):
+
+    images = Dict(
+        key_type=TextLine(),
+        value_type=TextLine(),
+        required=False,
+        default={}
+    )
+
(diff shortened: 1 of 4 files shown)


guillotina/tests/test_content.py
```diff
@@ -9,13 +9,34 @@
...
 import pytest
 
 
+class ICustomContentType(IItem):
+
+    images = Dict(
+        key_type=TextLine(),
+        value_type=TextLine(),
+        required=False,
+        default={}
```
bloodbare: Should be with factory, otherwise is the global object and the test does not makes sense

masipcat: I ran the tests without the fix and I could reproduce the bug (that's because `_default_from_schema()` creates a deepcopy of the default value: https://github.com/plone/guillotina/blob/master/guillotina/content.py#L138). I think this implementation details should be documented to know when to use `default` and `defaultFactory`. That said, I have no problem to change this :)

masipcat:
@bloodbare @vangheem I updated the changelog with a patch version but I'm not sure... If you think this should be a minor or major change, I'll change it :-)

bloodbare:
@masipcat You will need to add also requirements to aiohttp 3.4.4, tests are red due to aiohttp update, on 3.5 AbstractRouter resolve is incompatible with guillotina right now.

@vangheem

setup.py
```diff
@@ -49,7 +49,7 @@
     package_data={'': ['*.txt', '*.rst', 'guillotina/documentation/meta/*.json']},
     packages=find_packages(),
     install_requires=[
-        'aiohttp>=3.0.0,<4.0.0',
+        'aiohttp==3.4.4',
```
vangheem: Can you change to `>=3.0.0,<3.5.0`?

Issue #410, opened 2019-01-25
Unresponsive guillotina after losing connectivity with postgres


masipcat:
I'm getting the following error after restarting the postgres:

```
Error on execution of view 343bd8bedf114f8fa853ea44dc5f4388
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/guillotina/traversal.py", line 261, in handler
    await commit(request, warn=False)
  File "/usr/local/lib/python3.7/site-packages/guillotina/transactions.py", line 22, in commit
    await get_tm(request).commit(request)
  File "/usr/local/lib/python3.7/site-packages/guillotina/db/transaction_manager.py", line 99, in commit
    return await shield(self._commit(request=request, txn=txn))
  File "/usr/local/lib/python3.7/site-packages/guillotina/db/transaction_manager.py", line 108, in _commit
    await txn.commit()
  File "/usr/local/lib/python3.7/site-packages/guillotina/db/transaction.py", line 304, in commit
    return await self._commit()
  File "/usr/local/lib/python3.7/site-packages/guillotina/db/transaction.py", line 320, in _commit
    await self.tpc_commit()
  File "/usr/local/lib/python3.7/site-packages/guillotina/db/transaction.py", line 375, in tpc_commit
    await self._strategy.tpc_commit()
  File "/usr/local/lib/python3.7/site-packages/guillotina/db/strategies/resolve.py", line 56, in tpc_commit
    await self.retrieve_tid()
  File "/usr/local/lib/python3.7/site-packages/guillotina/db/strategies/base.py", line 29, in retrieve_tid
    tid = await self._storage.get_next_tid(self._transaction)
  File "/usr/local/lib/python3.7/site-packages/guillotina/db/storages/pg.py", line 761, in get_next_tid
    return await self._stmt_next_tid.fetchval()
  File "/usr/local/lib/python3.7/site-packages/asyncpg/connresource.py", line 19, in _check
    self._check_conn_validity(meth.__name__)
  File "/usr/local/lib/python3.7/site-packages/asyncpg/prepared_stmt.py", line 223, in _check_conn_validity
    super()._check_conn_validity(meth_name)
  File "/usr/local/lib/python3.7/site-packages/asyncpg/connresource.py", line 44, in _check_conn_validity
    self.__class__.__name__, meth_name))
asyncpg.exceptions._base.InterfaceError: cannot call PreparedStatement.fetchval(): the underlying connection is closed
```

[...]

PR #411, opened 2019-01-28
Improved _check_bad_connection()

masipcat:
Closes https://github.com/plone/guillotina/issues/410

There are more places where we should add this `except asyncpg.exceptions.PostgresConnectionError`. Maybe we could use a decorator like `@restart_conn_on_exception` that calls `_check_bad_connections()` for all the except cases.

### DIFF
--- a/guillotina/db/storages/pg.py
+++ b/guillotina/db/storages/pg.py
@@ -235,6 +235,20 @@ RETURNING id;
 ''')
 
 
+def restart_conn_on_exception(f):
+    async def decorator(self: 'PostgresqlStorage', *args, **kwargs):
+        try:
+            return await f(self, *args, **kwargs)
+        except asyncpg.exceptions.PostgresConnectionError as ex:
+            await self._check_bad_connection(ex)
+            raise
+        except asyncpg.exceptions.InterfaceError as ex:
+            await self._check_bad_connection(ex)
+            raise
+
+    return decorator
+
+
 # how long to wait before trying to recover bad connections
 BAD_CONNECTION_RESTART_DELAY = 0.25
 
@@ -646,13 +660,10 @@ WHERE tc.constraint_name = '{}_parent_id_id_key' AND tc.constraint_type = 'UNIQU
             await conn.execute("DROP TABLE IF EXISTS {};".format(self._blobs_table_name))
             await conn.execute("DROP TABLE IF EXISTS {};".format(self._objects_table_name))
 
+    @restart_conn_on_exception
     async def open(self):
-        try:
-            conn = await self.pool.acquire(timeout=self._conn_acquire_timeout)
-            return conn
-        except asyncpg.exceptions.InterfaceError as ex:
-            async with self.lock:
-                await self._check_bad_connection(ex)
+        conn = await self.pool.acquire(timeout=self._conn_acquire_timeout)
+        return conn
 
     async def close(self, con):
         try:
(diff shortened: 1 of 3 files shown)


guillotina/db/storages/pg.py
```diff
@@ -235,6 +235,20 @@ def _wrap_return_count(txt):
...
 
 
+def restart_conn_on_exception(f):
+    async def decorator(self: 'PostgresqlStorage', *args, **kwargs):
+        try:
+            return await f(self, *args, **kwargs)
+        except asyncpg.exceptions.PostgresConnectionError as ex:
+            await self._check_bad_connection(ex)
+            raise
+        except asyncpg.exceptions.InterfaceError as ex:
```
vangheem: I think this can be combined with above right?

`except (asyncpg.exceptions.PostgresConnectionError, asyncpg.exceptions.InterfaceError) as ex`

masipcat: Ups, you're right!

vangheem:
Do you have any idea what is wrong here with CR adapter? https://travis-ci.org/plone/guillotina/jobs/485546794

guillotina/tests/test_postgres.py
```diff
@@ -117,6 +118,32 @@
...
+    tm = TransactionManager(aps)
+
+    # Check 'read_conn' works
+    await tm._storage.get_next_tid(Mock())
+
+    # Simulate connection closed
+    tm._storage._connection_manager._read_conn.terminate()
+
+    # Don't know why we need this line but without this sleep fails
+    await asyncio.sleep(1)
```
masipcat: @vangheem do you know why this test fails without this line?

vangheem: how does it fail?

masipcat: Without the `sleep(1)` I get this error:

```python
    @pytest.mark.skipif(DATABASE == 'DUMMY', reason='Not for dummy db')
    async def test_restart_connection_pg(db, dummy_request):
        request = dummy_request  # noqa so magically get_current_request can find
    
        aps = await get_aps(db)
        tm = TransactionManager(aps)
    
        # Check 'read_conn' works
        await tm._storage.get_next_tid(Mock())
    
        # Simulate connection closed
        tm._storage._connection_manager._read_conn.terminate()
    
        # Don't know why we need this line but without this sleep fails
        #await asyncio.sleep(1)
    
        with pytest.raises(ConflictError):
>           await tm._storage.get_next_tid(Mock())

guillotina/tests/test_postgres.py:138: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
guillotina/db/storages/pg.py:241: in decorator
    return await f(self, *args, **kwargs)
guillotina/db/storages/pg.py:778: in get_next_tid
    return await self._stmt_next_tid.fetchval()
venv/lib/python3.6/site-packages/asyncpg/connresource.py:19: in _check
    self._check_conn_validity(meth.__name__)
venv/lib/python3.6/site-packages/asyncpg/prepared_stmt.py:223: in _check_conn_validity
    super()._check_conn_validity(meth_name)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <asyncpg.prepared_stmt.PreparedStatement object at 0x7fba53772548>, meth_name = 'fetchval'

    def _check_conn_validity(self, meth_name):
        con_release_ctr = self._connection._pool_release_ctr
        if con_release_ctr != self._con_release_ctr:
            raise exceptions.InterfaceError(
                'cannot call {}.{}(): '
                'the underlying connection has been released back '
                'to the pool'.format(self.__class__.__name__, meth_name))
    
        if self._connection.is_closed():
            raise exceptions.InterfaceError(
                'cannot call {}.{}(): '
                'the underlying connection is closed'.format(
>                   self.__class__.__name__, meth_name))
E           asyncpg.exceptions._base.InterfaceError: cannot call PreparedStatement.fetchval(): the underlying connection is closed

venv/lib/python3.6/site-packages/asyncpg/connresource.py:44: InterfaceError

```

vangheem: ahh, could it be that `terminate()` schedules a call later to handle the connection cleanup on a pool?

Without the sleep, the loop never schedule code to be called?

You could also try it with `sleep(0.0)` to see if it's just loop "context" switching

vangheem: maybe https://github.com/MagicStack/asyncpg/blob/master/asyncpg/connection.py#L1808 ?

vangheem: I wonder if there is a better way to simulate the connect being severed...

masipcat: > You could also try it with sleep(0.0) to see if it's just loop "context" switching

sleep(0) doesn't work :(

I checked the code tagged v0.17.0 and the only suspicious code I see is this method: https://github.com/MagicStack/asyncpg/blob/v0.17.0/asyncpg/connection.py#L1103

> I wonder if there is a better way to simulate the connect being severed...

Before implementing this test I tried to restart the pg container to simulate a real disconnect, but this would require a persistent volume and don't know how could be done while running in travis.

I'm open to hear a better way to simulate this behavior.

vangheem: You could patch and manually throw the error?

masipcat: After patching the method manually (and removing the sleep) the test continues failing. I put some breakpoints and found that the problem is here: https://github.com/plone/guillotina/blob/master/guillotina/db/storages/pg.py#L760 (the constant has a value of `0.25`).

masipcat: Btw, I think its confusing that `@restart_conn_on_exception` sometimes raises `ConflictError` (https://github.com/plone/guillotina/blob/master/guillotina/db/storages/pg.py#L564) and sometimes the real exception (`https://github.com/plone/guillotina/pull/411/files#diff-d558ee25390b1019470ae02f3f71861bR245`). Does it make sense to keep this behavior?

vangheem: yah, blah. Raising ConflictError is probably because we want to try to retry the request.

This is really messy.

masipcat: Yes... Well, this PR is ready for me

PR #420, opened 2019-02-08
Adding = to valid chars

bloodbare:


vangheem:
uh oh, we need to disable mr.roboto here.

vangheem:
@bloodbare why `=`? I mean, I guess it's fine but an `=` is also used for query params so it'll never look nice in a url.

vangheem:
and changelog!

bloodbare:
I want to store base64 urls

vangheem:
Wwwaaaahh? Haha, ok :)

PR #422, opened 2019-02-12
fixing write_permission bug

lferran:


vangheem:
@gforcada can we disable these mr roboto warnings for this project?

gforcada:
Yes, I will do that.

Issue #424, opened 2019-02-12
Add option to choose postgresql schema


masipcat:
I'd like to add the option `db_schema` to postgresql storage. The idea is to run guillotina on a database that already has other tables, but I want to isolate guillotina tables in a postgres schema (https://www.postgresql.org/docs/11/ddl-schemas.html I ). I tried using custom table names (like `object_table_name: guillotina.objects`) but guillotina fails when creates indices:

statement:`CREATE INDEX IF NOT EXISTS guillotina.objects_tid ON guillotina.objects (tid);` 
error: `asyncpg.exceptions.PostgresSyntaxError: syntax error at or near "."`

My idea is to add the option `db_schema` and fix the indices name when this option is defined.

I'll open a PR with a change proposal.

vangheem:
@masipcat This sounds great; however, I will warn you that I think I've tried something similar and I think asyncpg might fight you a little.

That being said, I might have been doing something wrong when I was playing with it.

Please let me know if you have trouble!

masipcat:
@vangheem I read that asyncpg fails when you change the schema (in the middle of a transaction) and then execute a cached statement.

I played with those changes locally and haven't found any problem so far. I'll try it in a dev environment during this week and I'll ping you with the result.

Thank you!

PR #425, opened 2019-02-12

masipcat:
Closes https://github.com/plone/guillotina/issues/424

This is the easiest approach I found but there are some alternative to this implementation:

1. Instead of storing `schema.table_name` in attribute `_objects_table_name` we can change all the sql statatements to accept the 'schema' as a parameter.
2. Execute `SET SCHEMA {schema}` when opening a connection
3. Ideas?

### DIFF
--- a/guillotina/db/storages/pg.py
+++ b/guillotina/db/storages/pg.py
@@ -13,6 +13,7 @@ from guillotina.db.interfaces import IPostgresStorage
 from guillotina.db.oid import MAX_OID_LENGTH
 from guillotina.db.storages.base import BaseStorage
 from guillotina.db.storages.utils import SQLStatements
+from guillotina.db.storages.utils import get_index_name
 from guillotina.db.storages.utils import get_table_definition
 from guillotina.db.storages.utils import register_sql
 from guillotina.exceptions import ConflictError
@@ -495,7 +496,7 @@ class PostgresqlStorage(BaseStorage):
 
     def __init__(self, dsn=None, partition=None, read_only=False, name=None,
                  pool_size=13, transaction_strategy='resolve_readcommitted',
-                 conn_acquire_timeout=20, cache_strategy='dummy',
+                 conn_acquire_timeout=20, cache_strategy='dummy', db_schema='public',
                  objects_table_name='objects', blobs_table_name='blobs',
                  connection_manager=None, **options):
         super(PostgresqlStorage, self).__init__(
@@ -510,8 +511,9 @@ class PostgresqlStorage(BaseStorage):
         self._options = options
         self._connection_options = {}
         self._connection_initialized_on = time.time()
-        self._objects_table_name = objects_table_name
-        self._blobs_table_name = blobs_table_name
+        self._db_schema = db_schema
+        self._objects_table_name = f'{db_schema}.{objects_table_name}'
+        self._blobs_table_name = f'{db_schema}.{blobs_table_name}'
         self._sql = SQLStatements()
         self._connection_manager = connection_manager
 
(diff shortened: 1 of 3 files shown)


guillotina/db/storages/pg.py
```diff
@@ -539,26 +546,26 @@ def lock(self):
...
+        statements.extend([
             get_table_definition(self._objects_table_name, self._object_schema),
             get_table_definition(self._blobs_table_name, self._blob_schema,
                                  primary_keys=('bid', 'zoid', 'chunk_index'))
-        ]
+        ])
         statements.extend(self._initialize_statements)
 
         for statement in statements:
-            otable_name = self._objects_table_name
```
vangheem: Unfortunately, this was a b/w compat way to not recreate indexes. It rewrites the default `objects` -> `object` and `blobs` -> `blob`.

So unfortunately, we still need something like this around... :(

guillotina/db/storages/pg.py
```diff
@@ -568,6 +575,18 @@ def lock(self):
 
         await self.initialize_tid_statements()
 
+    def _get_index_name(self, table_name):
+        if '.' in table_name:
+            _, table_name = table_name.split(".", 1)
+
+        if table_name in ("objects", "blobs"):
+            table_name = table_name[:-1]  # singluar index name
```
vangheem: Only for statements above ^^ that I commented on but not most!

Unless `_get_index_name` is only used for statement init?

guillotina/db/storages/pg.py
masipcat: I moved the logic here https://github.com/plone/guillotina/pull/425/files/edf11f986aefdf4139891c543bd2b414b7648316#diff-d558ee25390b1019470ae02f3f71861bR582

guillotina/db/storages/pg.py
masipcat: yes, `_get_index_name` is only used in the statement init

masipcat: I just moved the code inside a function to avoid duplicating the same logic (convert table_name to singular and append schema to index_name).

All those changes shouldn't affect the current behavior.

guillotina/db/storages/pg.py
```diff
@@ -568,6 +575,18 @@ def lock(self):
...
+
+        if self._db_schema is None:
+            return table_name
+        else:
+            return f'{self._db_schema}_{table_name}'
```
jordic: Thouhgt, there's no need to concat tablename to the index. (The index lands on the schema refered by the table)

masipcat: Hmm, in this case I'll remove this code and use previous implementation

masipcat:
@vangheem I think it makes sense to add again `_get_index_name()` because we need to clear the table_name in several places (some of them in addons: https://github.com/guillotinaweb/guillotina_elasticsearch/blob/master/guillotina_elasticsearch/commands/vacuum.py#L86)

vangheem:
@masipcat alright, I think it's fine then. Just be careful about how you customize it! :)

masipcat:
@jenkins-plone-org please run jobs

vangheem:
@masipcat oh, just ignore that plone robot!

masipcat:
@vangheem yeah, but after forced-push CI stopped working...

vangheem:
hmmm yah, travis.

Can you recreate PR? Or branch directly from plone/guillotina? I don't see your branch anywhere else in travis: https://travis-ci.org/plone/guillotina/builds

masipcat:
@vangheem btw, I'd like to run all the tests using a `db_schema: 'custom'` to ensure that everything still works after those changes. Should I use a env var like the one used here to choose which database (https://github.com/plone/guillotina/blob/master/guillotina/tests/fixtures.py#L25) ?

vangheem:
@masipcat yes, an additional env variable that configures the db differently seems fine to me!

PR #430, opened 2019-02-18

masipcat:
Closes https://github.com/plone/guillotina/issues/424


masipcat:
@vangheem somethin

PR #431, opened 2019-02-18

### LINKED ISSUE #424: Add option to choose postgresql schema

masipcat:
Closes #424


masipcat:
(https://github.com/plone/guillotina/pull/425)

@vangheem now it seems that work :-)

PR #435, opened 2019-03-07
Fix `GuillotinaDBRequester.make_request()` not decoding json responses

masipcat:
When a `Service` uses `json_response()` to response with JSON data, the content type is `Content-Type: application/json; charset=utf-8` and the previous implementation doesn't work. I think its better to designate the responsibility to detect JSON responses to aiohttp.

Thank you!


vangheem:
Thank you!

masipcat:
:-)

PR #436, opened 2019-03-08
Add UnionField type

vangheem:


vangheem:
@dmanchon elasticsearch allows single or array values by default. This is mostly to be more compatible with elasticsearch.

dmanchon:
Yeah, that case its clearly fine, but do we want to limit somehow someone creating an indexed field of type: UnionType(Int, Date)? because that will be impossible to map on elasticsearch.

vangheem:
well, for the UnionField implementation, it's fine; however, yes, we could add validation in guillotina_elasticsearch to handle this!

guillotina/api/search.py
```diff
@@ -116,7 +117,7 @@ def __init__(self, context, request, security=False):
 
 
 @configure.service(
-    context=IResource, method='POST',
+    context=IContainer, method='POST',
```
bloodbare: Why are we removing this feature, its good to be able to delete a catalog on a subresource

vangheem:
@bloodbare 

> What is the use case of UnionField?

https://docs.python.org/3/library/typing.html#typing.Union

guillotina/api/search.py
vangheem: Do we? I don't understand why we would want this?

bloodbare: Because the resource could be a subindex

guillotina/api/search.py
```diff
@@ -3,12 +3,13 @@
...
 from guillotina.component import query_utility
 from guillotina.interfaces import ICatalogUtility
+from guillotina.interfaces import IContainer
 from guillotina.interfaces import IResource
 from guillotina.utils import get_content_path
 
 
 @configure.service(
-    context=IResource, method='GET', permission='guillotina.SearchContent', name='@search',
+    context=IContainer, method='GET', permission='guillotina.SearchContent', name='@search',
```
bloodbare: Searching on a resource meaning that the resource is the basic path to search for.

guillotina/api/search.py
vangheem: alright, I guess

vangheem:
@bloodbare 

> Shouldnt we have a way to define a field on allowed types on directives.py:117 ?

you can already customize the index definition. Anything with index field definition is beyond the scope of this PR though. I'm not touching that ehre.

vangheem:
4.5.7 released

PR #438, opened 2019-03-11
Run `request.execute_futures` with managed_transaction context manager

vangheem:
and document

### DIFF
--- a/guillotina/transactions.py
+++ b/guillotina/transactions.py
@@ -17,6 +17,11 @@ def _safe_get_request(request):
 
 
 async def commit(request=None, warn=True):
+    '''
+    Commit the current active transaction.
+
+    :param request: request object transaction is connected to
+    '''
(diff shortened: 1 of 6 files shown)


guillotina/transactions.py
```diff
@@ -135,3 +167,6 @@ def adopt_objects(self, obs, txn):
                 self.request._txn = self.previous_txn
             if self.previous_write_setting is not None:
                 self.request._db_write_enabled = self.previous_write_setting
+
+        if self.request is not None:
```
bloodbare: Did you mean to check this.execute_futures ?
Should we sheck if request.execute_futures exist ?

vangheem: You are correct! Thank you for catching that.

"this.execute_futures": I can tell you've been doing a lot of JavaScript lately :)

PR #442, opened 2019-03-13
Provide vacuum command

vangheem:


guillotina/catalog/__init__.py
```diff
@@ -58,3 +60,13 @@ def get_parent_uuid(ob):
 @directives.index_field.with_accessor(IResource, 'tid', type='keyword')
 def get_tid(ob):
     return ob._p_serial
+
+
+@directives.index_field.with_accessor(
```
bloodbare: Is this needed allways, I mean on g_elasticsearch that would mean storing on all object field of the same container that information.

vangheem: good point, I moved to only jsonb

PR #446, opened 2019-03-20
Using symetric key on jwcrypto

bloodbare:


superit23:
Changing the algorithm in the config does not affect the algorithm used when the token is generated (`auth/__init__.py`) but only when it's validated (`auth/validators.py`). This causes all tokens to fail validation.
From `auth/__init__.py:43`
```python
jwt_token = jwt.encode(data, app_settings['jwt']['secret']).decode('utf-8')
```
From `auth/validators.py:99`
```python
validated_jwt = jwt.decode(
    token['token'],
    app_settings['jwt']['secret'],
    algorithms=[app_settings['jwt']['algorithm']])
```

bloodbare:
@superit23 the change is related to how temporal tokens are generated. JWE is used to issue a token that will be used on GET operations (like websocket) to send it as a GET parameter due to the restriction that websockets connections does not allow Headers and to avoid falling on the cookie based auth. This tokens are encrypted with a shared value so its not possible to introspect them as they are GET parameters.

The jwt library is used to issue the login token, a different token, to keep the session (that are only signed with the shared key) used on HEADERS based auth.

I don't understand your comment.

R

vangheem:
@bloodbare it's a comment about the jwt implementation. The `algorithm` setting isn't used for encoding.

The PR is great otherwise!

Issue #450, opened 2019-03-22
Implement mypy-zope


vangheem:
https://github.com/Shoobx/mypy-zope

PR #452, opened 2019-03-24
integrate mypy-zope

vangheem:
Fixes https://github.com/plone/guillotina/issues/450

### DIFF
--- a/guillotina/interfaces/json.py
+++ b/guillotina/interfaces/json.py
@@ -8,10 +8,10 @@ _ = MessageFactory('guillotina')
 class IResourceSerializeToJson(Interface):
     """Adapter to serialize a Resource into a JSON object."""
 
-    def __init__(self, context, request):
+    def __init__(context, request):
         """Adapt context and request."""
 
-    def __call__(self):
+    def __call__():
         """Return the json."""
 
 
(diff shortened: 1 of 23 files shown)


guillotina/interfaces/json.py
```diff
@@ -8,10 +8,10 @@
 class IResourceSerializeToJson(Interface):
     """Adapter to serialize a Resource into a JSON object."""
 
-    def __init__(self, context, request):
+    def __init__(context, request):
```
bloodbare: Whatt ?? no self ??

guillotina/schema/tests/test_schema.py
```diff
@@ -13,43 +13,47 @@
...
-            description="Description",
-            default=b"",
-            required=True)
-        spam = Bytes(
-            title="Spam",
-            description="Spam",
-            default=b"",
-            required=True)
-    return ISchemaTest
+    return InterfaceClass(
```
bloodbare: Do we need to change in all the interfaces defined inside code ???

guillotina/interfaces/json.py
vangheem: This is what way interfaces are supposed to be defined. zope's mypy supports complains about it.

guillotina/schema/tests/test_schema.py
vangheem: Just nested class defined interfaces might cause this it seems right now. mypy's integration throws an error on this. The reason is that it can't resolve the reference. Nested class defined interfaces don't have a proper `__module__`.  The `InterfaceClass` variant is identical and does proper `__module__`.

In my opinion, the `InterfaceClass` approach also looks nicer. I'm not a fan of using classes defined inside functions that use the function's closure.

bloodbare: 👍

Issue #463, opened 2019-04-03
Guillotina command not waiting for all aio tasks to finish


masipcat:
Hi,

I'm running latest Guillotina with Guillotina Rediscache. I created a custom command to generate/modify some guillotina objects. The code looks like this:

```python
class ModifyObj(Command):
    description = "Modify obj"

    async def run(self, arguments, settings, app):
        request = self.request
        aiotask_context.set("request", request)

        async for _, _, container in get_containers(request):
            await modify_objects()
            await request._txn.commit()
```

After runnning the command the other instances still have the old object. The problem is that the command ends without waiting for all aio tasks to finish (https://github.com/plone/guillotina/blob/master/guillotina/commands/__init__.py#L211) so `RedisCache._synchronize()` is never called (https://github.com/masipcat/guillotina_rediscache/blob/master/guillotina_rediscache/cache_strategy.py#L163).

Is there any reason to just wait for tasks with `qualname` == `Pool.release`?

masipcat:
Now I rewrote the previous code to:

```python
class ModifyObj(Command):
    description = "Modify obj"

    async def run(self, arguments, settings, app):
        request = self.request
        aiotask_context.set("request", request)

        async for _, _, container in get_containers(request):
            async with managed_transaction(request=request):
                await modify_objects_1()  # modifies a lot of objects
            # (implicit txn commit and close)

            async with managed_transaction(request=request):
                await modify_objects_2()  # modifies a lot of objects
            # (implicit txn commit and close)
```
But it seems that `managed_transaction()` isn't closing the transaction on `__aexit__()` so `RedisCache._synchronize()` never gets called.

masipcat:
@vangheem is this a bug ^ ?

vangheem:
I'm not sure what you mean about not closing the transaction.

It explicitly calls abort/commit https://github.com/plone/guillotina/blob/master/guillotina/transactions.py#L144

which should close the transaction..

If it is not, something must be wrong?

masipcat:
Ok, I continued debugging the problem and I found the root cause. This is a simplified version of `modify_objects()`:

```python
def modify_objects():
    request = get_current_request()
    container = request.container
    ob = await container.async_get("ob")
    ob.change_field = True
    ob._p_register()

    assert ob._p_jar == request._txn  # different!
```

[...]

PR #464, opened 2019-04-03
Fix Command.__run() not waiting for all aio tasks to finish

masipcat:
Closes https://github.com/plone/guillotina/issues/463

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,7 +1,8 @@
 4.5.12 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Fix Command.__run() not waiting for all aio tasks to finish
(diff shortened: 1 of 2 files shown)


masipcat:
> So this will wait up to 1 second for each task?

Yes, I just kept the previous implementation but maybe we can try with asyncio.gather on all tasks without a timeout.

Do you want me to change the implementation with asyncio.gather?

vangheem:
I think I'm okay leaving it. Could you maybe add a `logger.warning` if it exceeds the timeout? In the `asyncio.TimeoutError` exception handler?

masipcat:
@vangheem done

vangheem:
Thank you!

PR #467, opened 2019-04-04
create container util

vangheem:


guillotina/api/container.py
```diff
@@ -53,6 +55,40 @@ class DefaultGET(Service):
...
+
+    await parent.async_set(container_id, container)
+    await container.install()
+
+    if owner_id is None:
+        try:
+            owner_id = get_authenticated_user_id()
+        except RequestNotFound:
+            # fallback to root user as the owner
+            owner_id = 'root'
```
bloodbare: Could we prevent assiging owner if there is no owner defined  on argss?

vangheem: yup, that is fine with me!

PR #478, opened 2019-04-15
Util function to get a database object

bloodbare:


guillotina/utils/content.py
```diff
@@ -234,3 +233,22 @@ def get_object_url(ob: IResource,
...
+async def get_database(db_id, root=None):
+    if root is None:
+        root = get_utility(IApplication, name='root')
+
+    if db_id in root:
+        db = root[db_id]
+        if IDatabase.providedBy(db):
+            return db
+
+    if db_id in app_settings['storages']:
```
vangheem: I don't think this is correct. The storages setting is for dynamic storages.

I think you'd need to go through each dynamic config like it is done in `iter_databases`

PR #479, opened 2019-04-15
Proper content api login attribute

bloodbare:


guillotina/contentapi.py
```diff
@@ -8,25 +8,42 @@
...
+def logout(request):
+    request.security = None
+    request._cache_groups = {}
+    request._cache_user = None
+
 
 class ContentAPI:
 
-    def __init__(self, db):
+    def __init__(self, db, user=None):
```
vangheem: where is `user` being stored for use?

guillotina/contentapi.py
```diff
@@ -35,9 +52,11 @@ def __init__(self, db):
         except RequestNotFound:
             self._existing_request = None
         aiotask_context.set('request', self.request)
+        await login(self.request)
```
vangheem: Not passing user from content api?

PR #484, opened 2019-04-25
Fix being able to move content into itself

vangheem:


vangheem:
this isn't quite right yet here.

vangheem:
alright, this is ready now.

PR #491, opened 2019-05-02
Specific permission for uploading files

lferran:


vangheem:
I think you are right. I just had another thought--if you do base64 payload when patch a resource, it wouldn't use this permission.

lferran:
> I don't think is a good idea to define a new permission for modify content proposal. should not happen that you can add content to a field of the object (either when its a file) without the modify content permission. You could implement in your implementation a check with your specific permission on the FileManager.

ok! I will look into it thanks :)

PR #502, opened 2019-05-15
Cross role permissions issues

vangheem:
## Problem
Deny/Allow permissions assigned to multiple roles the user has are not properly applied.

This test demonstrates the issue.

I attempted to fix it only to realize there are other cases we won't cover.

So fixing it completely would introduce a lot of complexity I think.

Do you have any thoughts on it? Am I misunderstanding something?

## How it currently fails

- /db/container/folder [Allow: guillotina.DeleteContent to guillotina.Member]
- /db/container/folder/item [Deny: guillotina.DeleteContent to guillotina.Anonymous]

The roles used here can be changed with anything.

In this case the user will still be able to delete the item even though they have been specifically denied.

My PR fixes the above use-case only to ruin the reverse use case where allow is done on the child and deny done on the parent.

### DIFF
--- a/guillotina/auth/groups.py
+++ b/guillotina/auth/groups.py
@@ -1,5 +1,6 @@
 from guillotina import configure
 from guillotina.auth.users import GuillotinaUser
+from guillotina.interfaces import Allow
 from guillotina.interfaces import IGroups
 from guillotina.utils import get_current_request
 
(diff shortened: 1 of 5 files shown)


bloodbare:
This is a complex subject, what do you prioritize, Allowing or Denying? You need to choose one of them and right now its Allow first, could you explain the use case with what permission is set on each parent and which check fails ?

vangheem:
Updated the description to explain specifically what is wrong.

bloodbare:
I see your problem, guilllotina permission system does not create a matrix on permission check, it creates a flat list, so you can Deny Role-Permissions on the same role if you want to take that in effect. Trying to develop any security system that merges what guillotina.Anonymous and guillotina.Member to conclude that both needs to be applied on traversal check would end up with a less performance system. I recomend to add Deny to guillotina.DeleteContent onto guillotina.Member on that item.

vangheem:
Alright, I just wanted to identify/expose the issues.

I wonder if there is some way to prevent people from setting conflicting security configuration where you'd get unexpected behavior.

PR #516, opened 2019-05-28
Fix bug where non-async object subscribers were getting called twice

vangheem:


vangheem:
cockroach tests are getting so flaky....

PR #520, opened 2019-06-06
Make sure utility you are providing also provides the interface you

vangheem:


vangheem:
I know! Maybe simply because we used to do `@utility` for all these.

PR #528, opened 2019-06-14
HTTP Cache support

lferran:


guillotina/traversal.py
```diff
@@ -299,6 +299,11 @@ def __init__(self, resource, request, view):
         else:
             request.execute_futures('failure')
 
+        # Add extra headers to response if needed
+        extra_hdrs = getattr(self.view, '__extra_headers__', {})
```
bloodbare: instead of {} should be None

guillotina/test_package.py
```diff
@@ -213,6 +213,16 @@ def __init__(self, settings=None, loop=None):
     return request.matchdict
 
 
+@configure.service(
```
bloodbare: Can you add tests in case you answer a Response object and a Streamed response object ?

guillotina/traversal.py
```diff
@@ -299,6 +299,11 @@ def __init__(self, resource, request, view):
...
+        if extra_hdrs:
```
bloodbare: It would be great that it can be a function so we could provide a strong_cache() function and generate the headers.

guillotina/subscribers.py
```diff
@@ -28,3 +30,38 @@ def modified_object(obj, event):
...
     return await sitemanager.adapters.asubscribers((event.object, event), None)
+
+
+@configure.subscriber(
+    for_=IRequestFinishedEvent)
+async def add_http_cache_headers(event):
+    """This will add, if configured, the corresponding http cache headers
+    on the response of GET requests
+    """
+    if event.request.method != 'GET':
```
bloodbare: CORS are OPTIONS

guillotina/traversal.py
```diff
@@ -299,6 +300,9 @@ def __init__(self, resource, request, view):
         else:
             request.execute_futures('failure')
 
+        await notify(RequestFinishedEvent(
+            request, self.view, resp))
```
bloodbare: Maybe would be great to have also the resource

guillotina/subscribers.py
```diff
@@ -28,3 +30,38 @@ def modified_object(obj, event):
...
+        # Only cache responses from get requests
+        return
+
+    httpcache_settings = app_settings.get('http_cache', {})
```
bloodbare: Use a function to define default cache policy with resource, view  and request as parameters so we can create specific cache policies and merge with view overwrite configuration

vangheem:
@lferran What should we do about this? I don't think this should be worked on until we merge the asgi branch.

lferran:
yes! we can close it for now. We can rework it on the 6.x branch

Issue #458, opened 2019-03-26
Investigate a asgi based server


vangheem:
pros/cons

jordic:
Pros: 
- big move on python community (quark, django, starlette..). Interoperavility. Like wsgi on the pas
- http2 support throught hyper.
- Share middlewares, mount other services...
- separation of concerns. web framework vs web server
Cons:
- Maturity: asgi 2 to 3 (braking changes)

masipcat:
+1

cleder:
IIRC https://github.com/codelv/enaml-web does need an asgi server, so if there is any ambition to design an interface for guillotina in enaml this would be required (?)

vangheem:
We've moving to move to asgi as it also helps us get to http2 support

PR #529, opened 2019-06-15
ASGI Support

masipcat:
https://github.com/plone/guillotina/issues/458


masipcat:
Moved to https://github.com/plone/guillotina/pull/532

PR #532, opened 2019-06-15
ASGI Support

masipcat:
https://github.com/plone/guillotina/issues/458


dmanchon:
TODO:
- get rid of `starlette.websockets.WebSocket` dependency
- move `WebSocketSession` to `async_asgi_testclient`
- improve the IRequest interface
- `AsgiStreamReader` is not able to support HTTP2 should be reimplemented correctly

masipcat:
1. Can we replace this snippet with a Exception if a server (like daphne) is trying to send a request before doing a lifespan? I don't think we want to support a server that doesn't support ASGI at all.
https://github.com/plone/guillotina/blob/df15f8152750d76e3be1c61d2e5eea0e123622d8/guillotina/asgi.py#L111-L113 

2. https://github.com/plone/guillotina/blob/df15f8152750d76e3be1c61d2e5eea0e123622d8/guillotina/tests/test_attachment.py#L11-L15 Is this fixing the error in test_attachments.py that says `_blob.chunks == 1` (instead of 2)? Because I fixed this in https://github.com/plone/guillotina/pull/532/commits/309dce8ba4c99f523229aed6749ff5f9a23452d6, it was a bug in the `AsgiStreamReader`.

3. Right now I'm working on fixing/implementing the retry logic on HTTP conflict (just saying to avoid duplicating work). After that, I think that we just need to clean the code (flake8) and we are ready to ask for review.

What do you think @dmanchon ?

dmanchon:
@masipcat 

1. agree! lets remove that condition
2. yeah, i had same fix ready, but i found that we never really test in "real" conditions, where the data is comming in chunks, basically without using the async generator we never tested the `more_body` field being true.
3. nice, ithink that is the only test error left! i just commited the websocket session implementation based on the starlette one, so we can remove the dependency

guillotina/tests/fixtures.py
```diff
@@ -359,7 +359,9 @@ def __init__(self, request):
 async def _clear_dbs(app):
     root = app.root
     # make sure to completely clear db before carrying on...
-    async for db in iter_databases(root):
+    for _, db in root:
```
masipcat: @vangheem I had to change this because it was loading (and creating) the database defined in `settings['storages']['db']` during the cleanup:
```
[37]   /home/jordi/gits/guillotina/guillotina/asgi.py(129)__call__()
-> await self.shutdown()
[38]   /home/jordi/gits/guillotina/guillotina/asgi.py(142)shutdown()
-> await clean(self)
[39]   /home/jordi/gits/guillotina/guillotina/tests/fixtures.py(366)_clear_dbs()
-> if IPostgresStorage.providedBy(storage) or ICockroachStorage.providedBy(storage):
[40]   /home/jordi/gits/guillotina/guillotina/utils/content.py(238)iter_databases()
-> db = await factory.get_database(db_name)
[41]   /home/jordi/gits/guillotina/guillotina/db/factory.py(186)get_database()
-> self.app[name] = await apply_coroutine(factory, name, config)
[42]   /home/jordi/gits/guillotina/guillotina/utils/misc.py(120)apply_coroutine()
-> return await result
[43]   /home/jordi/gits/guillotina/guillotina/db/factory.py(84)PGDatabaseConfigurationFactory()
-> return await _PGConfigurationFactory(key, dbconfig, loop=loop)
[44]   /home/jordi/gits/guillotina/guillotina/db/factory.py(67)_PGConfigurationFactory()
-> aps = storage_factory(**dbconfig)
[45] > /home/jordi/gits/guillotina/guillotina/db/storages/pg.py(524)__init__()
-> self._dsn = dsn
(Pdb++) ll
 515         def __init__(self, dsn=None, partition=None, read_only=False, name=None,                   
 516                      pool_size=13, transaction_strategy='resolve_readcommitted',                   
 517                      conn_acquire_timeout=20, cache_strategy='dummy', db_schema='public',          
 518                      objects_table_name='objects', blobs_table_name='blobs',                       
 519                      connection_manager=None, autovacuum=True, **options):                         
 520             super(PostgresqlStorage, self).__init__(                                               
 521                 read_only, transaction_strategy=transaction_strategy,                              
 522                 cache_strategy=cache_strategy)                                                     
 523             import pdb; pdb.set_trace()                                                            
 524  ->         self._dsn = dsn                                                                        
 525             self._pool_size = pool_size                                                            
 526             self._partition_class = partition                                                      
 527             self._read_only = read_only                                                            
 528             self.__name__ = name                                                                   
 529             self._conn_acquire_timeout = conn_acquire_timeout                                      
 530             self._options = options                                                                
 531             self._connection_options = {}                                                          
 532             self._connection_initialized_on = time.time()                                          
 533             self._db_schema = db_schema                                                            
 534             self._objects_table_name = f'{db_schema}.{objects_table_name}'                         
 535             self._blobs_table_name = f'{db_schema}.{blobs_table_name}'                             
 536             self._sql = SQLStatements()                                                            
 537             self._connection_manager = connection_manager                                          
 538             self._autovacuum = autovacuum                                                          
```
After that, the `DELETE FROM` was crashing with `UndefinedTableError` for this table.

Any idea?

masipcat:
@dmanchon can you move the `WebSocketSession` to https://github.com/vinissimus/async-asgi-testclient, please? 

And after this, do you think we can mark the PR as ready for review?

guillotina/tests/fixtures.py
vangheem: Yes, you have it correct. We should not be doing iter_databases there! Good catch.

As for the `DELETE FROM`. Which one? How do I reproduce?

masipcat: You can reproduce it running this test (on commit 5269254c858796e930edec600655bd9023e2aba7):

`DB_SCHEMA=hola DATABASE=postgres pytest -s --cov=guillotina -s --tb=native -v --cov-report term-missing --cov-append guillotina/tests/test_annotations.py -x -s `

vangheem: I'm sorry, maybe I think I misunderstood. I can not reproduce the failure.

I thought you were saying the `DELETE FROM` was still a problem!

Thank you!

dmanchon:
> @dmanchon can you move the `WebSocketSession` to https://github.com/vinissimus/async-asgi-testclient, please?
> 
> And after this, do you think we can mark the PR as ready for review?

@masipcat i need acess to the async-asgi-testclient please

guillotina/tests/fixtures.py
masipcat: My bad, removing iter_databases solved the error.

Thank you!

masipcat:
> > @dmanchon can you move the `WebSocketSession` to https://github.com/vinissimus/async-asgi-testclient, please?
> > And after this, do you think we can mark the PR as ready for review?
> 
> @masipcat i need acess to the async-asgi-testclient please

Done!

bloodbare:
Do you mind adding the WIP label on the PR until its ready to be reviewed?

masipcat:
@bloodbare I think we are ready for a first review (all tests passing, we just need to add some more test to increase test coverage)

guillotina/asgi.py
```diff
@@ -0,0 +1,208 @@
...
+            self._buffer = rest
+            return req_chunk
+
+    async def receive(self):
+        payload = await self._asgi_receive()
+        self.finished = not payload.get("more_body", False)
+        return payload["body"]
+
+
+class AsgiStreamWriter():
```
masipcat: @dmanchon what do you think moving the writer to `response.py`?

guillotina/asgi.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+logger = glogging.getLogger('guillotina')
+
+
+def headers_to_list(headers):
+    return [[k.encode(), v.encode()] for k, v in headers.items()]
+
+
+class AsgiStreamReader(EmptyStreamReader):
```
masipcat: @dmanchon what do you think moving the reader to `request.py`? To avoid circular imports

masipcat:
> AsgiStreamReader is not able to support HTTP2 should be reimplemented correctly

@dmanchon do you know if this ^ is still failing?

guillotina/asgi.py
dmanchon: good idea!

guillotina/asgi.py
dmanchon: same, it makes sense to have all that logic together

dmanchon:
> > AsgiStreamReader is not able to support HTTP2 should be reimplemented correctly
> 
> @dmanchon do you know if this ^ is still failing?

that was when we have the dummy streamreader, now that is a proper implementation probably  will work fine! we should add tests tought

bloodbare:
We are planning to push ASGI for G6 so we have more time to test it

vangheem:
We are going to start releasing 5 alphas this week. It's a lot to put into a 5 release to have asgi right now.

I'm okay with quickly moving to asgi though after 5 is looking good. I'll also considering running 5 and 6 branches at the same time so we can make staying in sync easier.

We can work out a timeline as well!

masipcat:
Ok, no problem! Today I started testing this branch in our api in Vinissimus and most of the tests are passing (there are a lot of breaking changes! :P). I expect to start running this branch on our staging environment in the following weeks so I'll keep you informed about how is working.

> I'm okay with quickly moving to asgi though after 5 is looking good. I'll also considering running 5 and 6 branches at the same time so we can make staying in sync easier. 

This would be awesome!

Thank you

vangheem:
Added WIP until 5.x is stable and we're ready to cut 6.x

guillotina/commands/server.py
```diff
@@ -1,20 +1,6 @@
-from aiohttp import web
+import uvicorn
```
vangheem: So we default with uvicorn but I guess we'll need docs/info on how to run guillotina with uvicorn/etc by the normal startup command for those `uvicorn guillotina:app_module_thingy` if possible.

guillotina/request.py
```diff
@@ -1,19 +1,136 @@
-import time
-import uuid
+from aiohttp.web_ws import WSMessage
```
vangheem: So we are still depending on aiohttp. Is it necessary? Could we remove the dependency?

Or was this decision done to make migration easier?

Any thoughts on migrating these into core guillotina?

guillotina/tests/conftest.py
```diff
@@ -6,6 +6,9 @@
     'cockroachdb/cockroach', 'v2.1.6')
 
 
+images.configure('postgresql', version='10')
```
vangheem: good

guillotina/request.py
masipcat: I should be easy to copy all the WS models to `guillotina/request.py`. I think that the only thing that is strongly coupled with aiohttp is the routing/urlmatchinfo logic

dmanchon: yeah, i think this can be removed  quite easily

vangheem:
@masipcat I will deal with the rebase when we cut 5.x branch and are ready to merge this to master.

masipcat:
@vangheem I have some changes in my computer. I'll try to finish it by the end of the week. Are you planning to cut g5 during next week?

masipcat:
@vangheem All tests are passing!! The only big thing that its pending to be done is find another router to remove completely aiohttp as a dependency.

vangheem:
Woot!

I'll look into getting rid of aiohttp completely over the weekend if I have time.

I think it's important we do that before we merge though.

Once https://github.com/plone/guillotina/pull/651 is reviewed and merged, I will cut a final 5 release and then I will be okay merging this to master when it's ready.

masipcat:
> Once #651 is reviewed and merged, I will cut a final 5 release and then I will be okay merging this to master when it's ready.

Yay!

-------- 

Tomorrow i'll try to write some tests for Request/Response objects (that manages the ASGI logic).

guillotina/commands/server.py
masipcat: I think we have these options (options are written in order of preference (desc)):

1. Do somthing like `entrypoint = ServerCommand.run` in `guillotina/__init__.py` so any ASGI server can start guillotina with `G_CONFIG=./config.json uvicorn guillotina:entrypoint` (we can't pass custom arguments to our app when calling with uvicorn or any other server). With this approach the properties `host` and `port` won't be configurable from the `config.json`.
1. Hard-coding in the `ServerCommand` a list of ASGI servers, so we can launch the server with `g -c config.json serve --uvicorn` or `g -c config.json serve --hypercorn --host 0.0.0.0 --port 8080` (the `host` and `port` could be read from the config as well). But this would required do something like: 
```python
if arg == "--uvicorn":
    import uvicorn
    uvicorn.run(app, host=...)
elif arg == "--hypercorn":
    ...
```
3. Provide the functions `get_settings()` and `make_app(settings=)` and delegate to the user this logic. For example, by doing `app = make_app(settings=get_settings(file="config.json"))` in `app.py`, and then running the server with `uvicorn app:app` or by calling the server from code, like `if __name__ == "__main__": uvicorn.run(app, **params)` and running the file with `python app.py`

I'm not sure about the third option, could be tricky to implement it.

Thoughts?

vangheem: Can we do a combination of 1 and 2?

`#2` would include a default server to run with(say uvicorn) ootb but then you can also override it with maybe `--asgi-server=module.path.to.runner`.

`#2` would then depend on asgi providing a generic way to hook into for this--I'm not sure there isn't one ootb.

I think I would at the very least like an ootb implementation where the `serve` command worked with a default supported/tested implementation.

Thoughts?

masipcat: > #2 would include a default server to run with(say uvicorn) ootb but then you can also override it with maybe --asgi-server=module.path.to.runner.

The problem is there isn't a standard way to call an asgi server from code. 

* `uvicorn`:
```python
import uvicorn
uvicorn.run(app, host='0.0.0.0', port='0.0.0.0')
```
* `hypercorn`:
```python
# https://pgjones.gitlab.io/hypercorn/api_usage.html
from hypercorn.asyncio import serve
from hypercorn.config import Config

config = Config()
config.bind = ["localhost:8080"]  # As an example configuration setting
asyncio.run(serve(app, config))
```
* `daphne`:
```python
# Doesn't provide documentation on how to run the server from code
```

> I think I would at the very least like an ootb implementation where the serve command worked with a default supported/tested implementation.

I think we can provide an ootb implementation that works with `uvicorn` and `hypercorn`, and the user can select which server wants to run with `--asgi-server=hypercorn` (default to `uvicorn`). Something like:
```python
if arguments.asgi_server == "uvicorn":
    import uvicorn
    uvicorn.run(app, host=argument.host, port=argument.port)
elif arguments.asgi_server == "hypercorn":
    from hypercorn.asyncio import serve
    from hypercorn.config import Config
    config = Config()
    config.bind = [f"{argument.host}:{argument.port}"]
    asyncio.run(serve(app, config))
else:
    raise Exception(f"Server {arguments.asgi_server} not supported")
```

What do you think?

vangheem: Okay, I'm good with `--asgi-server`.

Maybe we can make sure to have good docs on creating app object to run with any server manually as well so users can integrate with another server before we have support for it. (or if they just want to have a different startup implementation)

masipcat: Done a first implementation. Suggestions are welcome :-)

masipcat:
I've done some benchmarks using `molotov`.


Using persisten db (`/df`) |   |   |  
-- | -- | -- | --
  | Measure 1 (35s) | Measure 2 (35s) | Avg req/s
(G5.0.12) Aiohttp+uvloop 3.5.4 | 6398 | 6785 | 188,33
Uvicorn 0.9.0 | 6806 | 6826 | 194,74
Hypercorn+uvloop 0.8.2 | 6289 | 6364 | 180,76


Using in-memory db (`/d`) |   |   |  
-- | -- | -- | --
  | Measure 1 (35s) | Measure 2 (35s) | Avg req/s
(G5.0.12) Aiohttp+uvloop 3.5.4 | 18921 | 18823 | 539,20
Uvicorn 0.9.0 | 20633 | 20655 | 589,83
Hypercorn+uvloop 0.8.2 | 16804 | 16801 | 480,07

I used this commands/configurations for the server:
* `uvicorn`: `G_CONFIG_FILE=config-dummy.yaml uvicorn --no-access-log guillotina.entrypoint:app`
* `hypercorn`: `G_CONFIG_FILE=config-dummy.yaml hypercorn guillotina.entrypoint:app -k uvloop`
* `daphne`: not supported (doesn't implement lifespan events)

and this command for `molotov`: `molotov molotov/test.py -p 6 -w 20 -d 35 -x --uvloop`.

vangheem:
Awesome! If you have time, can you compare that to Guillotina 4/5?

masipcat:
Duration: 60s
Command: `molotov molotov/test.py -p 8 -w 30 -d 60 -x --uvloop`

Db disk |   |   |     |  
-- | -- | -- | -- | --
  | Measure 1 | Measure 2 |    Avg req/s | Diff
G4+uvloop | 8497 | 8716 |    143,44 | 0,00%
G5+uvloop | 7090 | 6468 |    112,98 | -21,23%
G6 Uvicorn 0.9.0 | 8299 | 8162 |   137,18 | -4,37%
G6 Hypercorn+uvloop 0.8.2 | 8150 | 8161 |    135,93 | -5,24%


Db in memory |   |     |   |  
-- | -- | -- | --  | --
  | Measure 1 | Measure 2 |    Avg req/s | Diff
G4+uvloop | 33364 | 33121 |   554,04 | 0,00%
G5+uvloop | 30147 | 30462 |   505,08 | -8,84%
G6 Uvicorn 0.9.0 | 32291 | 33048 |   544,49 | -1,72%
G6 Hypercorn+uvloop 0.8.2 | 28297 | 28098 |  469,96 | -15,18%

masipcat:
image

I've been playing with pyflames this weekend and I think that we could improve the performance avoiding `yarl.URL`. Right now we are only using `request.url` in:

https://github.com/plone/guillotina/blob/d9a9de06ff8fd29ce3ebb55241b3c375b52b0608/guillotina/glogging.py#L26
https://github.com/plone/guillotina/blob/d9a9de06ff8fd29ce3ebb55241b3c375b52b0608/guillotina/utils/misc.py#L322

The asgi server already provides the parts of the url splitted (scheme, method, headers, raw_path, query_string) and the guillotina request converts the query_string and headers to multidict, therefore the only thing is missing is providing the full url (but should be as easy as concatenate all components).

What do you think about removing `yarl` @vangheem @bloodbare ?

Here is the svg if you want to check yourself: [out.txt](https://github.com/plone/guillotina/files/3648108/out.txt) (rename txt to svg)

bloodbare:
I like the concept of removing yarl, composing URLs on the get_url is used a lot, which is important to push performance there. We need to be able to provide our own path and switch the schema, do you think that manual replacement will be faster than yarl ?

masipcat:
> We need to be able to provide our own path and switch the schema, do you think that manual replacement will be faster than yarl ?

Well, right now we are constructing the yarl.URL twice. First time when we are calling `request.url` and second time in `get_url()` with `req.url.with_path(path)`. I'll try doing the concatenation with https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlunparse just one time and see if it's faster. I hope it will be faster :P I'll publish the results when it's done!

masipcat:
Updated benchmark after removing yarl:


Db disk |   |   |   |  
-- | -- | -- | -- | --
  | Measure 1 | Measure 2 | Avg req/s | Diff
G4+uvloop | 8497 | 8716 | 143,44 | 0,00%
G5+uvloop | 7090 | 6468 | 112,98 | -21,23%
G6 Uvicorn 0.9.0 | 8299 | 8162 | 137,18 | -4,37%
**G6 Uvicorn 0.9.0 (without yarl)** | **8759** | **8610** | **144,74** | **0,91%**
G6 Hypercorn+uvloop 0.8.2 | 8150 | 8161 | 135,93 | -5,24%


Db in memory |   |   |   |  
-- | -- | -- | -- | --
  | Measure 1 | Measure 2 | Avg req/s | Diff
G4+uvloop | 33364 | 33121 | 554,04 | 0,00%
G5+uvloop | 30147 | 30462 | 505,08 | -8,84%
G6 Uvicorn 0.9.0 | 32291 | 33048 | 544,49 | -1,72%
**G6 Uvicorn 0.9.0 (without yarl)** | **35548** | *+35009** | **582,68** | **5,17%**
G6 Hypercorn+uvloop 0.8.2 | 28297 | 28098 | 469,96 | -15,18%

vangheem:
Awesome job!

masipcat:
I'm ready for a review @bloodbare @vangheem ! and I think this is ready as a first alpha

vangheem:
thank you! I will review tonight.

Just to be clear and make sure everyone is on the same page:
- cut 5.x branch
- merge this to master
- cut alpha release

@bloodbare @jordic We ready to for this?

requirements.txt
```diff
@@ -14,6 +12,9 @@ python-dateutil==2.6.1
 PyYaml==5.1
 six==1.11.0
 ujson==1.35
-yarl==1.2.6
 zope.interface==4.4.3
+# Force build uvloop from source instead of using wheel
+# Replace with a uvicorn version that uses uvloop built for python 3.8
+git+https://github.com/dmontagu/uvloop@patch-1
```
masipcat: ^ Temporary fix. I guess this would be fixed in the following days.

bloodbare:
@masipcat that holds the merge until uvloops merge is done?

masipcat:
@bloodbare The problem with uvloop is that the official release is built with Cython <0.29.12 (doesn't support python 3.8) and is distributed with binaries (python wheels). Building uvloop from source using Cython 0.29.13 works without any other change.

So, I don't think we need to wait, but it's up to you ;)

guillotina/asgi.py
```diff
@@ -0,0 +1,118 @@
...
+        except Exception as e:
+            logger.exception("Something crashed during app startup")
+            raise e
+
+    async def shutdown(self):
+        for clean in self.on_cleanup:
+            await clean(self)
+
+    async def handler(self, scope, receive, send):
+        if not self.initialized:
```
bloodbare: Maybe a lock/queue in case there is lot of traffic and waits that its initialized?

requirements.txt
bloodbare: That may delay the merge

docs/source/training/extending/websockets.md
```diff
@@ -25,7 +23,9 @@ logger = logging.getLogger('guillotina_chat')
     context=IContainer, method='GET',
     permission='guillotina.AccessContent', name='@conversate')
 async def ws_conversate(context, request):
-    ws = web.WebSocketResponse()
+    ws = request.get_ws()
+    await ws.prepare(request)
```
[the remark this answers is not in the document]
masipcat: Hmm... I'll take a look

guillotina/asgi.py
masipcat: It's not needed. An ASGI server that implements the lifespan events will take care to call the init before accepting requests and calling the handler. This conditional was added because the ASGI server `daphne` doesn't implement the lifespan events (https://github.com/django/daphne/issues/264) so we can fail with a proper error.

I'll add a comment to the code.

docs/source/training/extending/websockets.md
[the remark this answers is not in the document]
masipcat: done

guillotina/asgi.py
masipcat: done

vangheem:
Thoughts on `git+https://github.com/dmontagu/uvloop@patch-1`? Maybe not depend on 3.8 yet?

masipcat:
I removed python 3.8 on travis and removed the `git+https...` in the requirements.

masipcat:
TODO: find the best way to make these tests work, beucase the async_asgi_testclient doesn't start a http server to run the tests, it calls the app directly... 

https://github.com/plone/guillotina/blob/2ed25be381279d91df70acbdd9254eace17db132/guillotina/tests/test_server.py#L47-L56

masipcat:
@bloodbare @vangheem all tests passing again.

TODO:
* investigate if uvicorn/hypercorn reuse asyncio task (with http1.1 + keep alive)

masipcat:
> @bloodbare @vangheem all tests passing again.
> 
> TODO:
> 
>     * investigate if uvicorn/hypercorn reuse asyncio task (with http1.1 + keep alive)

I checked uvicorn and hypercorn implemetantions and I found that both are creating a task for each request, as you can see here: https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/http/h11_impl.py#L234. Also, I think the `H11Protocol` is called for each request (or for each session if it has keep_alive) so, it reuses the task of the H11Protocol but creates a task for each request.

vangheem:
@bloodbare I'm about to push the merge button after tests pass here.

@masipcat is it okay if I squash all the commits?

masipcat:
> @masipcat is it okay if I squash all the commits?

@vangheem yes!

bloodbare:
💪👏

PR #535, opened 2019-06-16
PG Catalog implementation

vangheem:
Still some work but would like an initial review as well

### DIFF
--- a/guillotina/contrib/catalog/pg.py
+++ b/guillotina/contrib/catalog/pg.py
@@ -434,7 +434,7 @@ class PGSearchUtility(DefaultSearchUtility):
         select_fields = ['id', 'zoid', 'json']
         arg_index = 1
         for idx, select in enumerate(query['selects']):
-            select_fields.append(sqlq(select.format(arg=arg_index)))
+            select_fields.append(select.format(arg=arg_index))
             sql_arguments.append(query['selects_arguments'][idx])
             arg_index += 1
 
--- a/guillotina/contrib/catalog/pg.py
+++ b/guillotina/contrib/catalog/pg.py
@@ -1,5 +1,4 @@
 import json
-from guillotina.const import TRASHED_ID
 import logging
 import typing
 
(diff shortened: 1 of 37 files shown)


vangheem:
Ready for review!

bloodbare:
Maybe we need to clean up guillotina_cms and do a performance test to make sure pgcatalog how is affecting standard REST opretaions

vangheem:
Yes. G cms needs updating and I'll also do performance tests. Should be same as before though. Although, if you were using store_json before without indexes, it'll likely be slower. I'll try to test but it might be hard to detect until you have a lot of data.

vangheem:
Also, elastic search package should be updated with parser from g cms along with adding it's own default raw parser I think.

guillotina/contrib/catalog/pg.py
```diff
@@ -0,0 +1,549 @@
...
+
+    @property
+    def idx_name(self) -> str:
+        return 'idx_objects_{}'.format(self.name)
+
+    @property
+    def index_sql(self) -> typing.List[str]:
+        return [
+            f'''CREATE INDEX CONCURRENTLY IF NOT EXISTS {sqlq(self.idx_name)}
+                ON objects ((json->>'{sqlq(self.name)}'));''',
```
masipcat: I think this should be `CREATE INDEX ... ON {storage._objects_table_name} (...)` (because of supporting custom postgres schemas).

The reason the tests are passing right now even when using `DB_SCHEMA=custom` is because of the function `_clear_dbs()`.
https://github.com/plone/guillotina/blob/a64bc8da76ca71e7c967bd9cd57b51f4bddb1117/guillotina/tests/fixtures.py#L284-L293

This function calls `iter_databases()` and during the process of iterating databases creates a new db 'postgres'. That's because `factory.get_names()` is returning `['postgres', 'guillotina']` (why 'postgres'?) and because 'postgres' is not in `loaded`, `factory.get_database(db_name)` creates the tables `(public.)objects` and `blobs`.

https://github.com/plone/guillotina/blob/4c621a6b782569c9cf23d90325e827491951bdef/guillotina/utils/content.py#L231-L240

You can make tests fail changing `_clear_dbs()` with the following code:
```python
async def _clear_dbs(root):
    # make sure to completely clear db before carrying on...
    for _, db in root:
        if not IDatabase.providedBy(db):
            continue
        storage = db.storage
        ...
```
But i'm not sure if there is a root problem in `iter_databases()` or somewhere else...

PR #537, opened 2019-06-17
Emit events for registry configuration changes

vangheem:


bloodbare:
Shouldnt we backport this to G5 ?

PR #544, opened 2019-06-20
decorator for schema validation

karannaoh:


guillotina/api/service.py
```diff
@@ -36,6 +40,23 @@ def __setattr__(self, name, value):
 
 
 class Service(View):
+    async def validate(self, parameters):
+        data = await self.request.json()
+        schema = parameters[0]['schema']['$ref'].split('/')[-1]
```
vangheem: Let's validate this in a more safe way. This can quite easily cause errors right now!

Additionally, let's get the schema and make sure it is correct *before* we even get here(when building the service in the `configure/__init__.py` file)

guillotina/configure/__init__.py
```diff
@@ -447,6 +447,15 @@ def __call__(self, func):
                 )
             # create new class with customizations
             klass = type(func.__name__, (func,), dict(func.__dict__))
+            klass.config = self.config
+            __parameters__ = klass.config.get('parameters', {})
+            if klass.config.get('validate', False):
```
vangheem: Get the schema here and validate it is correct before moving on.

karannaoh: yep

guillotina/configure/__init__.py
```diff
@@ -447,6 +447,15 @@ def __call__(self, func):
                 )
             # create new class with customizations
             klass = type(func.__name__, (func,), dict(func.__dict__))
+            klass.config = self.config
+            __parameters__ = klass.config.get('parameters', {})
```
vangheem: Instead of `__parameters__` here, just use `__config__` instead.

guillotina/configure/__init__.py
```diff
@@ -447,6 +447,15 @@ def __call__(self, func):
...
+            if klass.config.get('validate', False):
+                original = klass.__call__
+
+                async def new_call(self):
+                    await self.validate(__parameters__)
```
vangheem: Let's not pass anything into the validate method.

Use the class `__config__` property.

guillotina/configure/__init__.py
```diff
@@ -469,9 +478,14 @@ def __call__(self, func):
             class _View(self.config.get('base', Service)):
                 __allow_access__ = self.config.get('allow_access', False)
                 __route__ = routes.Route(self.config.get('name', ''))
+                __auto_validate__ = self.config.get('validate', False)
+                __parameters__ = self.config.get('parameters', {})
```
vangheem: remove this line then.

guillotina/configure/__init__.py
```diff
@@ -469,9 +478,14 @@ def __call__(self, func):
...
+                __config__ = self.config
                 view_func = staticmethod(func)
 
                 async def __call__(self):
+                    if self.__auto_validate__:
+                        await self.validate(self.__parameters__)
```
vangheem: again, remove `self.__parameters__`

guillotina/api/service.py
```diff
@@ -36,6 +40,23 @@ def __setattr__(self, name, value):
 
 
 class Service(View):
+    async def validate(self, parameters):
```
vangheem: Do not pass in `parameters` here. Use `self.__config__`

guillotina/api/registry.py
```diff
@@ -98,21 +98,13 @@ class Read(Service):
     context=IContainer, method='POST',
     permission='guillotina.RegisterConfigurations', name='@registry',
     summary='Register a new interface to for registry settings',
+    validate=True,
     parameters=[{
```
vangheem: Reminder, in open api 3, describing request bodies changed: https://swagger.io/docs/specification/describing-request-body/

guillotina/api/service.py
```diff
@@ -36,6 +42,30 @@ def __setattr__(self, name, value):
 
 
 class Service(View):
+    __validator__ = None
+
+    async def validate(self):
+        data = await self.request.json()
+        if(self.__class__.__validator__ is None):
```
vangheem: `self.__class__` should just be `self` right?

guillotina/api/service.py
```diff
@@ -36,6 +42,30 @@ def __setattr__(self, name, value):
...
+            try:
+                self.__class__.__schema__ = self.__config__['parameters'][0]['schema']['$ref'].split('/')[-1]
+                self.__class__.__validator__ = get_schema_validator(self.__class__.__schema__)
+            except Exception as e:
+                logger.warning('Invalid jsonschema: {}'.format(e))
```
vangheem: Can we just iterate over the parameters and inspect the config to get the values correct instead of having a blind except block here?

guillotina/configure/__init__.py
```diff
@@ -447,6 +447,14 @@ def __call__(self, func):
                 )
             # create new class with customizations
             klass = type(func.__name__, (func,), dict(func.__dict__))
+            klass.__config__ = self.config
```
vangheem: Looks good here.

guillotina/api/service.py
```diff
@@ -36,6 +42,30 @@ def __setattr__(self, name, value):
 
 
 class Service(View):
+    __validator__ = None
+
+    async def validate(self):
+        data = await self.request.json()
+        if(self.__class__.__validator__ is None):
+            try:
+                self.__class__.__schema__ = self.__config__['parameters'][0]['schema']['$ref'].split('/')[-1]
```
vangheem: Also, this needs to look at open api 3's way of describing request body: https://swagger.io/docs/specification/describing-request-body/

guillotina/api/registry.py
karannaoh: for OpenAPI change I am working on another PR(https://github.com/plone/guillotina/pull/515/files). you want me put all the change for OpenAPI in this PR only?

vangheem: Yes, let's merge put these together since they depend on one another now.

guillotina/api/service.py
```diff
@@ -36,6 +42,31 @@ def __setattr__(self, name, value):
...
+    __validator__ = None
+
+    async def validate(self):
+        data = await self.request.json()
+        if(self.__validator__ is None):
+            try:
+                requestBody = self.__config__['requestBody']
+                self.__schema__ = requestBody['content']['application/json']['schema']['$ref'].split('/')[-1]
+                self.__validator__ = get_schema_validator(self.__schema__)
+            except Exception as e:
```
vangheem: Maybe we only catch `(IndexError, KeyError)` here?

guillotina/api/service.py
```diff
@@ -36,6 +42,31 @@ def __setattr__(self, name, value):
...
+                logger.warning('Invalid jsonschema: {}'.format(e))
```
vangheem: If there is an error, how about we set `self.__validator__` to a sentinel value like `_sentinel = object()` so we won't continue to output the warning + retry calculating schema?

I can show you this if this doesn't make sense.

karannaoh:
> Looking much better. Can we update `validate` to work with arbitrary json schemas, not just ones with a `$ref`?
> 
> Also, can we move getting the validator into it's own method? Maybe name it `_get_validator`?

guillotina/api/service.py
karannaoh: yep, I will add validation for arbitrary schemas and add a method for getting validator.

guillotina/api/service.py
karannaoh: Do you mean a flag kind of thing? `if self.__validator__ = "invalidschema"` then if wont give warning or retry calculating schema?

vangheem: Close. Take a look at how aiohttp does it.

- https://github.com/aio-libs/aiohttp/blob/master/aiohttp/helpers.py#L81
- https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L196

guillotina/api/service.py
```diff
@@ -36,6 +42,52 @@ def __setattr__(self, name, value):
...
+                    logger.warning(e)
+            else:
+                logger.warning("No schema found in service definition")
+        else:
+            pass  # can be used for query, path or header parameters
+        return [schema, validator]
+
+    async def validate(self):
+        data = await self.request.json()
+        if(self.__validator__ is None and self.__validator__ != self._sentinal):
```
vangheem: Move to `_get_validator`

guillotina/api/service.py
```diff
@@ -36,6 +42,52 @@ def __setattr__(self, name, value):
...
+                except jsonschema.exceptions.ValidationError as e:
+                    logger.warning(e)
+            else:
+                logger.warning("No schema found in service definition")
+        else:
+            pass  # can be used for query, path or header parameters
+        return [schema, validator]
+
+    async def validate(self):
+        data = await self.request.json()
```
vangheem: Move to just before validate call.

vangheem: Move to line 78 after `if self.__validator__`

guillotina/api/service.py
```diff
@@ -36,6 +42,52 @@ def __setattr__(self, name, value):
...
+                    validator = get_schema_validator(schema_name)
+                except KeyError as e:
+                    logger.warning('Invalid jsonschema: {}'.format(e))
+            elif requestBody['content']['application/json']['schema'] is not None:
+                try:
+                    schema = requestBody['content']['application/json']['schema']
+                    jsonschema_validator = jsonschema.validators.validator_for(schema)
+                    validator = jsonschema_validator(schema)
+                except jsonschema.exceptions.ValidationError as e:
+                    logger.warning(e)
```
vangheem: can be changed to `logger.warning('Could not validate schema', exc_info=True)`

guillotina/api/service.py
```diff
@@ -36,6 +42,52 @@ def __setattr__(self, name, value):
...
+        validator = self._sentinal
+        if config['requestBody']:
+            requestBody = config['requestBody']
+            if '$ref' in requestBody['content']['application/json']['schema']:
+                try:
+                    schema_name = requestBody['content']['application/json']['schema']['$ref'].split('/')[-1]
+                    schema = app_settings['json_schema_definitions'][schema_name]
+                    validator = get_schema_validator(schema_name)
+                except KeyError as e:
+                    logger.warning('Invalid jsonschema: {}'.format(e))
```
vangheem: Use `logger.warning('Invalid jsonschema, exc_info=True)` instead

guillotina/api/service.py
```diff
@@ -36,6 +42,56 @@ def __setattr__(self, name, value):
 
 
 class Service(View):
+    __validator__ = None
+    _sentinal = object()
+
+    def _get_validator(self, config):
+        [schema, validator] = [None, None]
```
vangheem: rewrite to `schema = validator = None`

guillotina/api/service.py
```diff
@@ -36,6 +42,56 @@ def __setattr__(self, name, value):
...
+        if(self.__validator__ is None and self.__validator__ != self._sentinal):
```
vangheem: remove enclosed `(` `)`

guillotina/api/service.py
```diff
@@ -36,6 +42,56 @@ def __setattr__(self, name, value):
...
+                        schema_name = ref.split('/')[-1]
+                        schema = app_settings['json_schema_definitions'][schema_name]
+                        validator = get_schema_validator(schema_name)
+                    except KeyError:
+                        logger.warning('Invalid jsonschema', exc_info=True)
+                elif requestBody['content']['application/json']['schema'] is not None:
+                    try:
+                        schema = requestBody['content']['application/json']['schema']
+                        jsonschema_validator = jsonschema.validators.validator_for(schema)
+                        validator = jsonschema_validator(schema)
```
vangheem: change to `self.__validator__ = jsonschema_validator(schema)`

guillotina/api/service.py
```diff
@@ -36,6 +42,56 @@ def __setattr__(self, name, value):
...
+        if(self.__validator__ is None and self.__validator__ != self._sentinal):
+            validator = self._sentinal
+            if config['requestBody']:
+                requestBody = config['requestBody']
+                if '$ref' in requestBody['content']['application/json']['schema']:
+                    try:
+                        ref = requestBody['content']['application/json']['schema']['$ref']
+                        schema_name = ref.split('/')[-1]
+                        schema = app_settings['json_schema_definitions'][schema_name]
+                        validator = get_schema_validator(schema_name)
```
vangheem: change to `self.__validator__ = get_schema_validator(schema_name)`

guillotina/api/service.py
```diff
@@ -36,6 +42,56 @@ def __setattr__(self, name, value):
...
+                        schema = requestBody['content']['application/json']['schema']
+                        jsonschema_validator = jsonschema.validators.validator_for(schema)
+                        validator = jsonschema_validator(schema)
+                    except jsonschema.exceptions.ValidationError:
+                        logger.warning('Could not validate schema', exc_info=True)
+                else:
+                    logger.warning("No schema found in service definition")
+            else:
+                pass  # can be used for query, path or header parameters
+            return [schema, validator]
```
vangheem: Remove next 3 lines and just replace with `return self.__validator__`

guillotina/api/service.py
```diff
@@ -36,6 +42,56 @@ def __setattr__(self, name, value):
...
+        else:
+            return [self.__schema__, self.__validator__]
+
+    async def validate(self):
+
+        [self.__schema__, self.__validator__] = self._get_validator(self.__config__)
```
vangheem: Just change to `validator = self._get_validator(self.__config__)`

and then make next line be `if validator is not None:`

guillotina/api/service.py
```diff
@@ -36,6 +42,56 @@ def __setattr__(self, name, value):
 
 
 class Service(View):
+    __validator__ = None
+    _sentinal = object()
+
+    def _get_validator(self, config):
```
vangheem: should we add type annotations?

karannaoh: I am sorry, What is annotation?

karannaoh: yep

karannaoh: `def _get_validator(self, config = None):`

guillotina/api/service.py
```diff
@@ -36,6 +42,53 @@ def __setattr__(self, name, value):
...
+                        self.__schema__ = requestBody['content']['application/json']['schema']
+                        jsonschema_validator = jsonschema.validators.validator_for(self.__schema__)
+                        self.__validator__ = jsonschema_validator(self.__schema__)
+                    except jsonschema.exceptions.ValidationError:
+                        logger.warning('Could not validate schema', exc_info=True)
+                else:
+                    logger.warning("No schema found in service definition")
+            else:
+                pass  # can be used for query, path or header parameters
+        return [self.__schema__, self.__validator__]
```
vangheem: change `[self.__schema__, self.__validator__]` to `self.__schema__, self.__validator__`

guillotina/api/service.py
```diff
@@ -36,6 +42,53 @@ def __setattr__(self, name, value):
...
+
+    async def validate(self):
+
+        [schema, validator] = self._get_validator()
```
vangheem: change `[schema, validator]` to `schema, validator`

guillotina/api/service.py
```diff
@@ -36,6 +42,53 @@ def __setattr__(self, name, value):
 
 
 class Service(View):
+    __validator__ = None
```
vangheem: Add default value for `__schema__`

guillotina/api/service.py
```diff
@@ -36,6 +42,53 @@ def __setattr__(self, name, value):
...
+    _sentinal = object()
+
+    def _get_validator(self):
```
vangheem: Change to being a `@classmethod`.

We want to cache this on the class itself, not the instance of it.

See https://www.programiz.com/python-programming/methods/built-in/classmethod for details on usage.

I think you had something close to this before and I forgot to redirect you to using a `@classmethod`. I think before you were doing something like `self.__class__.__schema__` -- what you really want to do is use classmethod.

guillotina/api/service.py
```diff
@@ -36,6 +42,53 @@ def __setattr__(self, name, value):
...
+                pass  # can be used for query, path or header parameters
+        return [self.__schema__, self.__validator__]
+
+    async def validate(self):
+
+        [schema, validator] = self._get_validator()
+        if validator:
+            try:
+                data = await self.request.json()
+                self.__validator__.validate(data)
```
vangheem: If you're returning `validator`, use it instead of `self.__validator__` here.

karannaoh:
> yay! I'm so excited we have swagger 3 now!

same !! :P

PR #546, opened 2019-06-20
fix interface for get_data

vangheem:


vangheem:
closing for https://github.com/plone/guillotina/pull/547

PR #550, opened 2019-06-21
add argon2 pw hashing

vangheem:


jordic:
I'm not pretty sure, is this called on the thread pool. This is cpu bound...

```
In [5]: %time ph.hash("123412341234")                                                                                                                                                                                
CPU times: user 277 ms, sys: 60.2 ms, total: 337 ms
Wall time: 142 ms
Out[5]: '$argon2id$v=19$m=102400,t=2,p=8$k6HLUbEcypi9dskGVCa6xw$C6dKsKmaqWho1PZdrxZthQ'
```
If not, seems like could be something dangerous.. (Just someone hitting the logging endpoint)

vangheem:
Alright, maybe we need to think about a different way to solve it.

salted SHA512 isn't ideal.

vangheem:
or were you thinking we should move it to `run_in_executor`?

jordic:
running the validators on a thread pool?

vangheem:
Yes, we can do that. A specific thread pool with limited workers I think.

jordic:
1 is enought, just to not be able to block the main thread.

vangheem:
@jordic updated. Can you have a look?

PR #551, opened 2019-06-21
Merging pubsub and rediscache

bloodbare:


guillotina/contrib/cache/__init__.py
```diff
@@ -0,0 +1,25 @@
+from guillotina import configure
+
+
+CACHE_PREFIX = 'gcache2-'
+
+app_settings = {
+    'cache': {
+        'driver': 'guillotina.contrib.redis',  # Empty to use memory
```
vangheem: Should we have default be empty then?

guillotina/contrib/cache/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+        'updates_channel': 'guillotina',  # Empty to not subscribe
```
vangheem: same here?

guillotina/contrib/cache/utility.py
```diff
@@ -0,0 +1,223 @@
...
+
+        assert isinstance(msg, dict)
+        assert 'tid' in msg
+        assert 'keys' in msg
+        if msg['tid'] in self._ignored_tids:
+            # on the same thread, ignore this sucker...
+            self._ignored_tids.remove(msg['tid'])
+            return
+
+        mem_cache_obj = memcache.get_memory_cache()
```
vangheem: should we use `self._memory_cache` instead?

guillotina/permissions.py
```diff
@@ -173,6 +174,9 @@
 configure.grant(
     permission="guillotina.RawSearchContent",
     role="guillotina.ContainerAdmin")
+configure.grant(
```
vangheem: Should this be defined in `guillotina.contrib.cache`?

guillotina/contrib/pubsub/__init__.py
```diff
@@ -0,0 +1,16 @@
+
+app_settings = {
```
vangheem: Should we add dependency of `guillotina.contrib.cache`?

`"applications": ["guillotina.contrib.cache"]`?

guillotina/contrib/pubsub/utility.py
```diff
@@ -0,0 +1,106 @@
+from guillotina.profile import profilable
+from guillotina.utils import resolve_dotted_name
+from guillotina.contrib.pubsub.exceptions import NoPubSubDriver
+from typing import Any
+from typing import Callable
+import pickle
+import asyncio
+import logging
+
+app_settings = {
```
vangheem: Why are these app_settings here? Shouldn't it be loading from the app_settings in parent module?

guillotina/contrib/redis/__init__.py
```diff
@@ -0,0 +1,30 @@
+from guillotina.contrib.redis.driver import RedisDriver
+from asyncio import get_running_loop
+_driver = None
+
+app_settings = {
```
vangheem: Same here about dependency maybe?

guillotina/contrib/cache/__init__.py
bloodbare: Yups

guillotina/contrib/cache/__init__.py
bloodbare: yups

guillotina/contrib/cache/utility.py
bloodbare: yups

PR #555, opened 2019-06-22
Make sure you can import contrib packages without automatically activating them

vangheem:
Should fix things for us


vangheem:
what would you prefer?

I don't like the magic where modules are automatically included without you noticing....

vangheem:
See https://github.com/plone/guillotina/issues/556 for proposal in future....

PR #563, opened 2019-06-24
be able to customize object reader

vangheem:
With `object_reader` setting


guillotina/contrib/redis/dm.py
```diff
@@ -34,6 +34,7 @@ class RedisFileDataManager(DBDataManager):
                 self._data = json.loads(data)
 
     async def start(self):
+        print('start')
```
bloodbare: Print!!!!

guillotina/contrib/redis/dm.py
```diff
@@ -42,6 +43,7 @@ class RedisFileDataManager(DBDataManager):
         txn.add_after_commit_hook(self._save)
 
     async def _save(self):
+        print('save')
```
bloodbare: Print!!!

guillotina/contrib/redis/dm.py
vangheem: woops, thought I fixed that!

PR #566, opened 2019-06-24
Integrate guillotina_dbusers 

jordic:


docs/source/contrib/dbusers.md
```diff
@@ -0,0 +1,59 @@
+
+# Guillotina dbusers
+
+Store users/groups in the database for guillotina.
+
+
+## Installation
+
+- add `guillotina_dbusers` to list of applications in your guillotina configuration
```
vangheem: no longer `guillotina_dbusers` here

guillotina/contrib/dbusers/services/groups.py
```diff
@@ -0,0 +1,81 @@
...
+from guillotina.interfaces import IContainer
+from guillotina.interfaces import IResourceSerializeToJsonSummary
+from guillotina.response import HTTPNotFound
+from guillotina.utils import navigate_to
+from zope.interface import alsoProvides
+
+
+@configure.service(
+    for_=IContainer, method="GET",
+    name="@groups"
```
bloodbare: Should not be a permission?

guillotina/contrib/dbusers/services/groups.py
```diff
@@ -0,0 +1,81 @@
...
+                (group, self.request), IResourceSerializeToJsonSummary
+            )
+            result.append(await serializer())
+
+        return result
+
+    async def get_groups_from_folder(self):
+        group_folder = await navigate_to(self.context, "groups")
+        groups = []
+        async for _, group in group_folder.async_items():
```
bloodbare: Should we do a search?

guillotina/contrib/dbusers/services/groups.py
```diff
@@ -0,0 +1,81 @@
...
+        group_id = self.request.matchdict["group"]
+        group = await navigate_to(self.context, "groups/{}".format(group_id))
+        if not group:
+            raise HTTPNotFound()
+        return group
+
+
+@configure.service(
+    for_=IContainer, methods="GET",
+    name="@groups/{group}"
```
bloodbare: Permission?

guillotina/contrib/dbusers/services/groups.py
```diff
@@ -0,0 +1,81 @@
...
+        group = await self.get_group()
+        serializer = get_multi_adapter(
+            (group, self.request), IResourceSerializeToJsonSummary
+        )
+        return await serializer()
+
+
+@configure.service(
+    for_=IContainer, method="PATCH",
+    name="@groups/{group}"
```
bloodbare: Permission

guillotina/contrib/dbusers/services/groups.py
```diff
@@ -0,0 +1,81 @@
...
+class PatchGroups(BaseGroup):
+    async def __call__(self):
+        group = await self.get_group()
+        alsoProvides(self.request, IPATCH)
+        view = DefaultPATCH(group, self.request)
+        return await view()
+
+
+@configure.service(
+    for_=IContainer, method="DELETE",
```
bloodbare: Permission

guillotina/contrib/dbusers/services/groups.py
jordic: It started with a search, but later, @vangheem removed it https://github.com/plone/guillotina_dbusers/commit/e3d84fe1c72a6df796aa2ecb0a550a9277e9a5e8

vangheem: Yes, sorry. get_by_type was removed. Do you think we need it?

vangheem: Can simply be done with .search(container, {type_name:})

Sorry, traveling so syntax is not perfect

jordic: Is the catalog installed by default?

vangheem: No, you will need to check if it is configured before using

bloodbare: Agree with .search

guillotina/contrib/dbusers/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+    "auth_user_identifiers": [
+        "guillotina.contrib.dbusers.users.DBUserIdentifier"
+    ]
+}
+
+configure.permission("guillotina.NotAuthenticated", "")
+configure.permission("guillotina.Authenticated", "")
+configure.permission("guillotina.AddUser", title="Add plone user")
+configure.permission("guillotina.AddGroup", title="Add plone group")
+configure.grant(permission="guillotina.AccessContent", role="guillotina.Anonymous")
```
lferran: isn't that kinda dangerous?

guillotina/contrib/dbusers/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+configure.grant(permission="guillotina.NotAuthenticated", role="guillotina.Anonymous")
+configure.grant(permission="guillotina.Authenticated", role="guillotina.Authenticated")
```
lferran: Everywhere else in the core,`guillotina.Authenticated` refers to the role.

Could this be confusing?

I guess if we would be using different prefixes for roles and permissions, it would help. Something like: `guillotina.permission:AccessContent` and `guillotina.role:Authenticated`

Maybe something worth doing it in the mid term?

guillotina/contrib/dbusers/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+configure.grant(permission="guillotina.AddUser", role="guillotina.ContainerAdmin")
+configure.grant(permission="guillotina.AddGroup", role="guillotina.ContainerAdmin")
+
```
lferran: Only principals with `guillotina.ContainerAdmin` role will be able to add new groups/users.

At the same time, the `guillotina.ManageUsers` permission is only given to `guillotina.Manager`

is that what we want? Shouldn't at least `guillotina.Manager` have `guillotina.AddUser` and `guillotina.AddGroup` permission?

guillotina/contrib/dbusers/content/groups.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+
+@configure.contenttype(
+    type_name="Group",
+    schema=IGroup,
+    add_permission="guillotina.AddGroup",
+    behaviors=["guillotina.behaviors.dublincore.IDublinCore"],
+)
+class Group(Folder):
```
lferran: I guess if `User` inherigs from `IPrincipal`, then `Group` should do aswell?

Do you remember why that was needed?

guillotina/contrib/dbusers/content/groups.py
```diff
@@ -0,0 +1,57 @@
...
+    pass
+
+
+class IGroup(IFolder):
+
+    name = schema.TextLine(title=_("Group name"), required=False)
+    description = schema.TextLine(title=_("Group Description"), required=False)
+    user_roles = schema.List(
+        title=_("Roles"), value_type=schema.TextLine(), required=False
+    )
```
lferran: Are those supposed to be actual guillotina global roles?
If so, don't we need to change the core security logic to conditionally (if dbusers is installed) check for this aswell?

guillotina/contrib/dbusers/content/users.py
```diff
@@ -0,0 +1,111 @@
...
+    @property
+    def permissions(self):
+        permissions = {}
+        for permission in getattr(self, 'user_permissions', []) or []:
+            permissions[permission] = Allow
+        return permissions
+
+    @property
+    def groups(self):
+        return getattr(self, 'user_groups', []) or []
```
lferran: Just for me to understand: are these properties in combination with `DBUserIdentifier` the magic that makes permissions and roles assigned to the `User` folder object be checked in core authorization logic?

guillotina/contrib/dbusers/content/users.py
```diff
@@ -0,0 +1,111 @@
...
+        title=_('Password'),
+        required=False)
+
+    user_groups = schema.List(
+        title=_('Groups'),
+        value_type=schema.TextLine(),
+        required=False
+    )
+
+    user_roles = schema.List(
```
lferran: I think this should also be write_protected so that a user cannot modify its own roles

guillotina/contrib/dbusers/content/users.py
```diff
@@ -0,0 +1,111 @@
...
+        title=_('Roles'),
+        value_type=schema.TextLine(),
+        required=False
+    )
+
+    user_permissions = schema.List(
```
lferran: TODO: add write_permission

guillotina/contrib/dbusers/content/users.py
```diff
@@ -0,0 +1,111 @@
...
+    name = schema.TextLine(
+        title=_('Name'),
+        required=False)
+
+    read_permission(password='guillotina.Nobody')
+    password = schema.TextLine(
+        title=_('Password'),
+        required=False)
+
+    user_groups = schema.List(
```
lferran: todo: add write_permission

guillotina/contrib/dbusers/__init__.py
vangheem: hmmm, yah, I wonder why that is there.

guillotina/contrib/dbusers/__init__.py
vangheem: I think we should see why these are here. Maybe we don't need any of these anymore. Maybe artifacts from old implementation?

guillotina/contrib/dbusers/__init__.py
vangheem: Yes, probably!

guillotina/contrib/dbusers/__init__.py
lferran: it seems it has been there from the beginning: https://github.com/plone/guillotina_dbusers/blame/739e0f9ada206402077894f62684f033f1200f7a/guillotina_dbusers/__init__.py#L18

guillotina/contrib/dbusers/content/users.py
bloodbare: Agree!

bloodbare:
@lferran thanks for taking care about it !

guillotina/contrib/dbusers/content/users.py
lferran: yup, confirmed!

guillotina/contrib/dbusers/__init__.py
lferran: I think I got to the root of the problem.
I fixed it with https://github.com/plone/guillotina/pull/566/commits/8e488e9b2886cc018bc084ebba0bf2534d884207

guillotina/contrib/dbusers/__init__.py
lferran: deleted them for now, as they seem unused... in case we need to add them back, I will add them properly

guillotina/contrib/dbusers/__init__.py
lferran: from the role description, I don't see why `ContainerAdmin` should have permissions to manage/create users.

Unless someone have reasons to do so, for now I gave those to `guillotina.Manager`

guillotina/contrib/dbusers/content/groups.py
lferran: not needed.

guillotina/contrib/dbusers/content/groups.py
lferran: not needed. 
Guillotina already checks permissions assigned to users' groups

guillotina/contrib/dbusers/__init__.py
vangheem: I think it makes sense for `ContainerAdmin` to have this access. `guillotina.Manager` is a very powerful user that usually only the root user has.

ContainerAdmin is more specific to managing containers or a specific container only.

guillotina/contrib/dbusers/content/groups.py
vangheem: But ferran, these are global roles a user gets assigned when authenticating against a container.

Think of this as what the roles an oauth/user api server would be returning for a user.

vangheem: Then again, I see your point. We could assign everything at the root of the container and get the same affect.

guillotina/contrib/dbusers/services/groups.py
bloodbare: @ferran can you try to load search utility and if its not then do this fallback ?

guillotina/contrib/dbusers/services/users.py
```diff
@@ -0,0 +1,139 @@
...
+        for user in users:
+            serializer = get_multi_adapter((user, self.request), IResourceSerializeToJsonSummary)
+            result.append(await serializer())
+
+        return result
+
+    async def get_users_form_folder(self):
+        user_folder = await navigate_to(self.context, "users")
+        users = []
+        async for _, user in user_folder.async_items():
```
bloodbare: Can we search on the catalog if its available or go on the folder as a fallback ?

guillotina/contrib/dbusers/users.py
```diff
@@ -0,0 +1,31 @@
...
+            users = await container.async_get("users")
+        except (AttributeError, KeyError, ContainerNotFound):
+            return
+
+        user_id = token.get("id", "")
+        if not user_id:
+            # No user id in the token
+            return
+
+        if user_id not in await users.async_keys():
```
bloodbare: This operation should be cached or a direct call to async_contains ?

guillotina/contrib/dbusers/services/groups.py
lferran: @bloodbare @vangheem any examples on a catalog search by type_name that works both elastic and pgcatalog?

vangheem: Yup, the new query catalog works in a generic way to parse queries!

just do...
```
await catalog.query(container, {"portal_type": "User"})
```

lferran: oh cool! thanks

lferran: I am now able to get users from the search.

Should I then get them individually from storage and output their summaryJSONSerialization?

lferran: fixed!

lferran: ready for re-review

lferran:
@bloodbare travis is failing due a weird mypy error. Tests are passing though.

guillotina/tests/dbusers/test_manage_groups.py
```diff
@@ -0,0 +1,95 @@
...
+
+        # ensure we cannot patch invalid users
+        data = {"users": {"foobarx": True}}
+        resp, status = await requester("PATCH", "/db/guillotina/@groups/foo", data=json.dumps(data))
+        assert status == 412
+
+
+settings_with_catalog = copy.deepcopy(settings.DEFAULT_SETTINGS)
+settings_with_catalog["applications"].append("guillotina.contrib.catalog.pg")
+settings_with_catalog.update(
```
vangheem: I think what you want to do here is:

`settings_with_catalog['load_utilities']['catalog'] = ...`

PR #567, opened 2019-06-25
Adding mailer

bloodbare:


guillotina/contrib/mailer/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+from guillotina.interfaces import IMailer
+from guillotina.component import provide_utility
+from guillotina.utils import import_class
+import logging
+
+logger = logging.getLogger('guillotina.contrib.mailer')
+
+app_settings = {
+    "mailer": {
+        "default_sender": "foo@bar.com",
```
vangheem: Should we leave `deault_sender` to None and log warning if not provided and trying to send?

guillotina/contrib/mailer/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+        "endpoints": {
+            "default": {
+                "type": "smtp",
```
vangheem: Same here I wonder?

guillotina/contrib/mailer/__init__.py
bloodbare: agree

guillotina/contrib/mailer/__init__.py
bloodbare: I would keep localhost as default option

bloodbare:
Yups, we should have it, can we plan to add on beta phase ?

PR #575, opened 2019-06-27
WIP Read only dsn

bloodbare:


vangheem:
If I understand this correctly, it's okay to close now! Please reopen if I'm wrong.

PR #580, opened 2019-07-02
Log exception on initialize utility

bloodbare:


guillotina/factory/content.py
```diff
@@ -66,8 +66,16 @@ def add_async_utility(self, key: str, config: typing.Dict,
             kw['name'] = config['name']
         provide_utility(utility_object, interface, **kw)
         if hasattr(utility_object, 'initialize'):
+            func = lazy_apply(utility_object.initialize, app=self.app)
+
+            async def notice_on_error(key, func_to_call):
```
vangheem: Should this be outside this scope?

PR #589, opened 2019-07-15
Read only txn are cached as read only

bloodbare:


guillotina/db/transaction.py
```diff
@@ -116,6 +100,25 @@ def __init__(self, manager, loop=None, read_only=False):
 
         self._query_count_start = self._query_count_end = 0
 
+    def clean(self, read_only):
```
vangheem: Maybe rename to `initialize` though?

PR #590, opened 2019-07-20
Allow to reindex on PG Catalog

bloodbare:


guillotina/contrib/catalog/pg.py
```diff
@@ -603,3 +619,62 @@ def load_meatdata(self, query: ParsedQueryInfo, data: typing.Dict[str, typing.An
         """
         recursively go through all content to reindex jsonb...
         """
+
+        self._sql = SQLStatements()
```
vangheem: You can also get this from the storage object as well

bloodbare:
Right now I'm more concern about development cycle. If you change global roles you need to reindex actual objects. We should add tests also to pg catalog code before adding queue processor.

PR #591, opened 2019-07-22
Allow to receive full object on search at pg catalog

bloodbare:


guillotina/contrib/catalog/pg.py
```diff
@@ -581,15 +583,28 @@ def load_meatdata(self, query: ParsedQueryInfo, data: typing.Dict[str, typing.An
...
             data = json.loads(record['json'])
-            result = self.load_meatdata(query, data)
-            result['@name'] = record['id']
-            result['@uid'] = record['zoid']
-            result['@id'] = data['@absolute_url'] = context_url + data['path']
+            if fullobjects and request is not None and txn is not None:
+                # Get Object
+                obj = await txn.get(data['uuid'])
+                # Serialize object
+                view = DefaultGET(obj, request)
```
vangheem: Maybe we should just call serializer on this? I'm not a fan of using view code as it could have side affects.

guillotina/contrib/catalog/pg.py
```diff
@@ -581,15 +583,28 @@ def load_meatdata(self, query: ParsedQueryInfo, data: typing.Dict[str, typing.An
         results = []
         try:
             context_url = get_object_url(container)
+            request = get_current_request()
+            fullobjects = request.query.get('_fullobjects', 'false').lower() != 'false'
```
vangheem: `_fullobjects` should be added to the query spec instead of looking at query param directly here.

here https://github.com/plone/guillotina/blob/master/guillotina/catalog/types.py#L6
or here https://github.com/plone/guillotina/blob/master/guillotina/contrib/catalog/pg.py#L71

guillotina/contrib/catalog/pg.py
bloodbare: I was not willing to update the logic of include and omit

guillotina/contrib/catalog/pg.py
bloodbare: Agreed, not willing to change the contract of the search endpoint so we need to update also ES

guillotina/contrib/catalog/pg.py
vangheem: ahh, so that include/omit can then also be used with these results

PR #596, opened 2019-07-23
[4.x] Make sure field name of patch field is set before using

vangheem:


bloodbare:
@vangheem should be also in 5?

vangheem:
@bloodbare yup, I'll make a PR there as well!

PR #599, opened 2019-07-24
[master] Fix exhausted retries conflict error response

vangheem:


vangheem:
exhausted http conflicts were causing 500 errors because we were provided guillotina response object instead of aiohttp response object.

so the key part here is: `from aiohttp.web_exceptions import HTTPConflict`

vangheem:
and the test wasn't caching this error properly. So yes, also update the test to test is it more correctly.

bloodbare:
Ouch missed with import movement! Great!

PR #601, opened 2019-07-24
[master] only register object to write if data changed

vangheem:


guillotina/json/deserialize_content.py
```diff
@@ -72,21 +72,22 @@ def __init__(self, context, request):
...
 
         return self.context
 
     async def set_schema(
             self, schema, obj, data, errors,
-            validate_all=False, behavior=False):
+            validate_all=False, behavior=False) -> bool:
         write_permissions = merged_tagged_value_dict(schema, write_permission.key)
-
+        changed = False
```
kakshay21: It's just a small concern to have that flag named `schema_changed` instead of `changed`. As it seems to give more context about the flag.

guillotina/json/deserialize_content.py
```diff
@@ -72,21 +72,22 @@ def __init__(self, context, request):
 
         factory = get_cached_factory(self.context.type_name)
         main_schema = factory.schema
-        await self.set_schema(
+        changed = await self.set_schema(
```
kakshay21: @vangheem IMO, `self.set_schema` method should not return anything(Ideally). Does it make sense to have this flag set in the `main_schema`  object and we can just check ?
```
if main_schema.schema_changed: self.context.register()
```

vangheem: The schema isn't changing though--the value on an instance object has changed.

`main_schema` here references a global reference to a schema interface(which is a python class). So `main_schema.schema_changed` wouldn't actually quite be appropriate.

I admit that having `set_schema` return a boolean feels awkward here though.

Also, naming... `set_schema` should probably be renamed to `set_values_from_schema`

kakshay21: yeah makes sense

vangheem:
@bloodbare sure but that's not the problem I'm trying to solve here and it's also not why we have this blind object register.

https://github.com/plone/guillotina/blob/master/guillotina/content.py#L211  registers the object so I don't think that problem is related.

Unless I misunderstand your comment.

guillotina/json/deserialize_content.py
vangheem: updated to not change return type

PR #608, opened 2019-07-30
Leak pg txn on reindex

bloodbare:


guillotina/contrib/catalog/pg.py
```diff
@@ -640,12 +640,18 @@ def load_meatdata(self, query: ParsedQueryInfo, data: typing.Dict[str, typing.An
...
+            return
+
+        data['transaction'] = await data['tm'].begin()
+        container.__txn__ = data['transaction']
         await self._process_object(container, data)
         await self._process_folder(container, data)
-        if data['transaction'] is not None:
-            await data['transaction'].commit()
-            data['transaction'] = None
+        await data['transaction'].commit()
```
vangheem: This should be `await data['tm'].commit(txn= data['transaction'])`

guillotina/contrib/catalog/pg.py
```diff
@@ -659,11 +665,9 @@ def load_meatdata(self, query: ParsedQueryInfo, data: typing.Dict[str, typing.An
 
     async def _process_object(self, obj, data):
 
-        if data['count'] % 2 == 0 and data['transaction'] is not None:
+        if data['count'] % 200 == 0:
             await data['transaction'].commit()
```
vangheem: same here

PR #609, opened 2019-07-31
bucket dict field implementation

vangheem:
Ended up with a sorted key/value approach so we can do splitting consistently.

Other approaches I tried and failed with:
- fixed bucket size with hashing(murmur2)
- unordered dynamically buckets with attempting to hash because of start/end key values--this worked extremely poorly

I want to also implement a `@fieldvalue` endpoint so we have a mechanism for iterating over these values. So draft for now but please review to yell if this doesn't look nice.

### DIFF
--- a/docs/source/developer/fields.md
+++ b/docs/source/developer/fields.md
@@ -318,6 +318,7 @@ from guillotina import schema
 
 class IMySchema(Interface):
     values = BucketDictField(
+        key_type=schema.Text(),
         value_type=schema.Text(),
         bucket_len=5000
     )
--- a/docs/source/developer/fields.md
+++ b/docs/source/developer/fields.md
@@ -23,6 +23,7 @@ consist of field definitions.
 - guillotina.schema.Time
 - guillotina.fields.PatchField: allow updating value without patching entire value
 - guillotina.fields.BucketListField: optimized storage for very large lists of data
+- guillotina.fields.BucketDictField: optimized storage for very large dictionaries of data
 - guillotina.files.CloudFileField: file field for storing in db or cloud storage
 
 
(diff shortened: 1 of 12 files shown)


guillotina/fields/annotation.py
```diff
@@ -133,43 +146,19 @@ def __init__(self, *args, value_type=None, bucket_len=5000, **kwargs):
...
-        'buckets': len(value.annotations_metadata)
-    }
-
-
 @configure.adapter(
     for_=IBucketListField,
     provides=IPatchFieldOperation,
     name='append')
 class PatchBucketListAppend(patch.PatchListAppend):
+    value_factory: typing.Any = BucketListValue
```
bloodbare: Why its needed ?

guillotina/fields/annotation.py
```diff
@@ -215,3 +204,299 @@ class PatchBucketListRemove(PatchBucketListAppend):
...
+                'created': time.time(),
+            }
+            self.buckets.insert(bidx + 1, new_bucket)
+            new_annotation = await self.get_annotation(context, middle_key)
+            # rebalance now
+            new_annotation['keys'] = annotation['keys'][middle_idx:]
+            new_annotation['values'] = annotation['values'][middle_idx:]
+            new_bucket['len'] = len(new_annotation['keys'])
+
+            annotation['keys'] = annotation['keys'][:middle_idx]
```
bloodbare: Del instead of copy

guillotina/fields/annotation.py
```diff
@@ -215,3 +204,299 @@ class PatchBucketListRemove(PatchBucketListAppend):
...
+            bidx, bucket = self._find_bucket(key)
+            middle_idx = int(self.bucket_len / 2)
+            middle_key = annotation['keys'][middle_idx]
+            new_bucket = {
+                'id': uuid.uuid4().hex,
+                'start': middle_key,
+                'created': time.time(),
+            }
+            self.buckets.insert(bidx + 1, new_bucket)
+            new_annotation = await self.get_annotation(context, middle_key)
```
bloodbare: New annotation register()

guillotina/fields/annotation.py
vangheem: Because it's generic for the dict field implementation

guillotina/fields/annotation.py
vangheem: Register is done in the `get_annotation` call

docs/source/developer/fields.md
```diff
@@ -306,3 +307,61 @@ Delete:
...
+## Bucket dict field
+
+```python
+from zope.interface import Interface
+from guillotina.fields import BucketDictField
+from guillotina import schema
+
+class IMySchema(Interface):
+    values = BucketDictField(
+        value_type=schema.Text(),
```
lferran: `key_type` is also usable with `BucketDictField` no?

lferran:
awesome... this is really useful!

docs/source/developer/fields.md
vangheem: I think I need to consider this. We need to force a key type that is consistently comparable when doing the bucket splitting...

So str, int, float, datetime would work.

Should we have checks for this or buyer beware or...?

PR #613, opened 2019-08-02
Dict serialization needs to be valid JSON schema

bloodbare:


bloodbare:
https://github.com/guillotinaweb/ngx-schema-form/issues/272

Issue #616, opened 2019-08-03
Swagger documentation for search


vangheem:
This is actually quite important and we're missing it. 

@karannaoh run guillotina with postgresql and have the `guillotina.contrib.catalog.pg` application activated. Add content and then use the `@search` endpoint: https://guillotina.readthedocs.io/en/latest/rest/container.html?highlight=search#search

@bloodbare we might need to improve our docs for activating/using pg catalog... Also see https://guillotina.readthedocs.io/en/latest/contrib/jsonbcatalog.html -- empty

PR #630, opened 2019-08-09
Swagger for search

karannaoh:


karannaoh:
#630

karannaoh:
#616

guillotina/api/service.py
```diff
@@ -45,11 +45,40 @@ class Service(View):
     __validator__ = __schema__ = None
     _sentinal = object()
 
+    def _validate_parameters(self):
+        if 'parameters' in self.__config__:
+            data = self.request.url.query
+            for parameter in self.__config__['parameters']:
+                if parameter['in'] == 'query':
+                    if parameter['schema']['type'] == 'integer':
```
vangheem: Could we handle KeyError here?

guillotina/api/service.py
```diff
@@ -45,11 +45,40 @@ class Service(View):
...
+                        try:
+                            int(data[parameter['name']])
```
vangheem: Before all this, shouldn't we first check if `name` is in `data`?

guillotina/api/service.py
```diff
@@ -45,11 +45,40 @@ class Service(View):
...
+                        try:
+                            float(data[parameter['name']])
+                        except ValueError:
+                            raise HTTPPreconditionFailed(content={
+                                'reason': 'Schema validation error',
+                                'message': 'can not convert {} to Float'.format(data[parameter['name']])
+                            })
+                    else:
+                        pass
+                    if parameter['required'] and parameter['name'] not in data:
```
vangheem: what if they don't define `required`?

Maybe the whole check should be wrapped in `KeyError` handling?

guillotina/api/service.py
karannaoh: keyError for parameter['schema']['type']?

vangheem: Yes, if they define a parameter without the `schema` defined

I imagine it already default to string so you don't need to define this.

Issue #631, opened 2019-08-13
Field `modification_date` is not updated in elasticsearch when doing a PATCH


masipcat:
This happens because the indexer uses the payload in the `ObjectModifiedEvent` to update only the fields that changed in ES.

https://github.com/plone/guillotina/blob/d3903c250562b3657973c2c8e7cb38c4ba407252/guillotina/api/content.py#L295

https://github.com/plone/guillotina/blob/0328970418e120796313cecab28a667c01249de5/guillotina/catalog/index.py#L135-L146

Adding `event.payload["modification_date"] = now.isoformat()` here will fix it:
https://github.com/plone/guillotina/blob/0328970418e120796313cecab28a667c01249de5/guillotina/subscribers.py#L14-L18
but i'm not sure if it's the best way...

What do you think? Thank you in advance!

vangheem:
I think it should be added to the default indexed data values:
here: https://github.com/plone/guillotina/blob/master/guillotina/catalog/catalog.py#L138

and here: https://github.com/plone/guillotina/blob/master/guillotina/catalog/catalog.py#L175

masipcat:
I knew there would be a better solution!

I'll prepare the PR

masipcat:
This was fixed in master https://github.com/plone/guillotina/pull/637. Closing

PR #632, opened 2019-08-13
[4.x] Fix `modification_date` not indexed when an object is patched

masipcat:
Closes https://github.com/plone/guillotina/issues/631


masipcat:
It was a couple of tests :P

masipcat:
`mypy-zope` was updated to 0.2.0 and this version requires `mypy` 0.720 (but was pinned 0.670)

vangheem:
4.9.3 released

PR #638, opened 2019-08-17
adding helm

karannaoh:


charts/guillotina/templates/service.yaml
```diff
@@ -0,0 +1,18 @@
...
+    run: {{ .Values.name }}
+spec:
+  ports:
+  - protocol: TCP
+    port: {{.Values.guillotinaPort}}
+    targetPort: {{.Values.guillotinaPort}}
+    nodePort: {{.Values.exposedPort}}
+  selector:
+    run: {{ .Values.name }}
+  type: NodePort
```
vangheem: Make this configurable

karannaoh:
> Also make sure that we have an easy way to customize the config.json
> 
> Otherwise, make all those settings be in values.yaml

@vangheem a person can just add a new config.json file according to his need, as we generally do while working. Along with it he as command line options too(which I guess you are asking) by passing an array of argument.

vangheem:
@karannaoh I would really love to have a good default config that is customizatable with helm chart values.

Having to provide your own config file completely makes the helm chart much less useful.

karannaoh:
> @karannaoh I would really love to have a good default config that is customizatable with helm chart values.
> 
> Having to provide your own config file completely makes the helm chart much less useful.

https://github.com/plone/guillotina/blob/0a02ce71a40f3408a0c63e42aa06692d734984c6/helm/guillotina/templates/configmap.yaml This is something I tried before(having config.yaml in configmap), I will define all the values inside values.yaml.

karannaoh:
@vangheem are there more changes you want me to work on?

charts/guillotina/values.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+serviceType: NodePort
+
+resources: {}
+
+guillotinaPort: 8080
+
+arguments: []
+
+configurations:
```
vangheem: Can we rename this to `app_settings`?

karannaoh: yep done

karannaoh:
@vangheem can we merge it?

vangheem:
I think it looks good but didn't we kind of decide that we were going to just work on it here but then the actual chart would go into helm's charts?

karannaoh:
> I think it looks good but didn't we kind of decide that we were going to just work on it here but then the actual chart would go into helm's charts?

I will create a PR to helm's charts repository. can we have one our own repository in guillotinaweb?

vangheem:
@karannaoh hmmm, if we think we want it in both places, maybe we should just merge this then? I'm okay with merging this--just need to know what the best course of action is here.

karannaoh:
@vangheem I tried to figure out what can be best.. I am seeing most of the projects maintain separate repositories for helm charts.

vangheem:
Let's just merge here. It'll be easier to maintain.

PR #646, opened 2019-08-24
Proposal: remove __txn__ from object

vangheem:
thoughts:

- `__txn__` feels magical
- adding objects to container always means using `parent.__txn__`
- in order to make this work most of the time, you need to have the magic in transaction manager which would reuse txn objects. Otherwise, parents wouldn't be able to add children across transaction boundaries without reassigning `__txn__` all the time. So it added weird cross dependencies
- conceptually, it's much easier to simply to `txn.add(ob)` -- explicit is better than implicit..

implications:
- maybe all object mutate api methods should accept a `txn` argument to be able to customize which txn to register it against(in case you are manually working with multiple transactions).
- I'd also like to find a nice way to auto retry conflict errors when manually working with txn objects. In practice, writing a generic api for this is difficult though.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,12 @@ CHANGELOG
 5.0.0a16 (unreleased)
 ---------------------
 
+- Remove IBaseObject.__txn__, IBaseObject.register, explicitly use `ITransaction.register` now
+  [vangheem]
+
+- No longer attempt to reuse a transaction object(unnecessary optimization)
+  [vangheem]
+
 - Fix cache key for SQLStatements cache. This was causing vacuuming on multi-db environments
   to not work since the vacuuming object was shared between dbs on guillotina_dynamictablestorage.
   [vangheem]
--- a/guillotina/annotations.py
+++ b/guillotina/annotations.py
@@ -3,8 +3,9 @@ from guillotina import configure
 from guillotina.db.orm.base import BaseObject
 from guillotina.interfaces import IAnnotationData
 from guillotina.interfaces import IAnnotations
-from guillotina.interfaces import IResource
 from guillotina.interfaces import IRegistry
+from guillotina.interfaces import IResource
+from guillotina.utils import get_current_transaction
 from zope.interface import implementer
 
 import logging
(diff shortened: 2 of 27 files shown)


bloodbare:
I think that removing reference from object to txn makes it difficult to have multiple transactions at the same time and adds complexity

bloodbare:
Besides we already moved all our code to G5 with txn reference and we sre using. That will make a lot of changes on our codebase (and needing to keep ref of txn and obj outside g)

vangheem:
@jordic @masipcat @lferran @bloodbare thoughts?

vangheem:
Alright, I'll write some demo code to demonstrate how ugly this API is right now.

vangheem:
fwiw, we're still alpha and we're able to still do decisions like this. A big point of the 5 release was to simply our txn/object machinery.

vangheem:
## What happens when we create an object...

- `await create_content_in_container(parent, 'Item')`
- builds object
- calls `await parent.async_set(obj.id, obj)`
- assigns txn from `parent.__txn__` and registers object with parent's txn


## So what happens if you commit and want to add to parent again?

- you would assume the parent no longer holds proper txn reference
- luckily, our transaction manager does a lookup of a old txn and reuses it if possible(maybe this isn't always possible)
- so even though the parent's txn was closed and finished, it magically gets reactivated


## What happens if this txn isn't able to be magically reactivated?

- your object will get registered with an invalid txn
- you will commit but your object would be edited
- Or, you *need* to keep reassigning the parents `.__txn__` value to a known good txn *every* time


This change will make our API feels like most other db APIs and be less magical.

@bloodbare I actually think this approach makes working with multiple transactions easier. Rather than always needing to set `.__txn__` on an object(which is very awkward), you just need to register the object with the txn you want.

We're still alpha so now is the time...

vangheem:
I'm not sure "I don't want to change my code" is a good reason to disapprove of this.

Let's talk about if this is better/right and then decide if it goes in g5 or g6 or whatever.

btw, sorry, @masipcat that we aren't on g6 yet... I'm also still sort of researching asgi. I will try and get g to final in the next few weeks and then cut  merge asgi to master with g6 alphas to come.

bloodbare:
I agree that reusing txn is bad so, a new transacion (after commit) should read again the parent

vangheem:
@bloodbare can I propose another thing then if this isn't going to be in g5?

How about if you attempt to register an object with a closed txn, we throw an error?

vangheem:
See https://github.com/plone/guillotina/pull/649

bloodbare:
Loving the proposal!

jordic:
Sorry! I was on holidays! Seems like there are two use cases:
The txn is managed implicit.. (there's no need to pass it explicit).. it's taken from the context var. If there is no current transaction active a new one is created. The txn is auto commited on request end.
The user is managing manually a transaction, then it should be passed explicit.
`parent.async_set(obj,.. txn=txn)`

vangheem:
@jordic we don't actually have auto commit type functionality and I'm not sure I want to do that :(

I prefer being explicit.

masipcat:
> @jordic @masipcat @lferran @bloodbare thoughts?

I like this proposal.

I remember some time ago when started working with guillotina and implemented some custom endpoints/commands that uses `managed_transaction()` inside the request to commit partial operations. My surprise was that  the objects I was registering inside the context manager of the nested transaction wasn't stored in pg until the parent transaction was committed. At the time I didn't found what was the problem but now I guess the problem was `obj._p_txn` was pointing to the parent txn instead of the nested one. Maybe this is not 100% accurate, I don't remember the details, but I've had more than one headache doing "advanced things" with transactions. I guess with the change you're proposing this wouldn't happen and therefore will be easier to work with transactions without knowing the internals of the framework.

PS: I'm on holidays as well :P

PR #651, opened 2019-08-27
G5 fixes

vangheem:
Closing https://github.com/plone/guillotina/pull/650
and instead going to have an ongoing PR for g5 issues

- Be able to configure cache to not push pickles with invalidation data
- Fix transaction handling to always get current active transaction, throw exception when transaction is closed and be able to refresh objects.
- More normalization of execute module with task_vars/request objects
- Allow committing objects that were created with different transaction
- Fix async utils to work correctly with transactions and context vars

### DIFF
--- a/guillotina/api/service.py
+++ b/guillotina/api/service.py
@@ -37,7 +37,9 @@ class DictFieldProxy:
             return super().__setattr__(name, value)
 
         if name == self.__field_name:
-            getattr(self.__context, name)[self.__key] = value
+            val = getattr(self.__context, name)
+            setattr(self.__context, name, val)
+            val[self.__key] = value
         else:
             setattr(self.__context, name, value)
 
--- a/guillotina/schema/utils.py
+++ b/guillotina/schema/utils.py
@@ -31,4 +31,6 @@ def get_default_from_schema(context, schema, fieldname, default=None):
             return deepcopy(field.defaultFactory())
     if field.default is not None:
         return deepcopy(field.default)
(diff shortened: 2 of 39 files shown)


guillotina/api/service.py
```diff
@@ -37,7 +37,9 @@ def __setattr__(self, name, value):
             return super().__setattr__(name, value)
 
         if name == self.__field_name:
-            getattr(self.__context, name)[self.__key] = value
+            val = getattr(self.__context, name)
+            setattr(self.__context, name, val)
```
bloodbare: Why do we need to set again the dict?

guillotina/schema/utils.py
```diff
@@ -31,4 +31,6 @@ def get_default_from_schema(context, schema, fieldname, default=None):
             return deepcopy(field.defaultFactory())
     if field.default is not None:
         return deepcopy(field.default)
+    if field.default != default:
```
bloodbare: If the field has `default = None` and we ask for another value we return `None`. Which use case covers this line?

vangheem: Right here: https://github.com/plone/guillotina/blob/master/guillotina/content.py#L252

guillotina/api/service.py
vangheem: Because of default field values.

This is used in the multi file behavior.

vangheem: basically, this was causing values to not be saved

PR #652, opened 2019-08-30
More Guillotina 5 fixes

vangheem:
- Fix json schema validation
- Fix memory cache to be able to calc size properly
- Better redis pubsub error handling


guillotina/contrib/pubsub/utility.py
```diff
@@ -41,27 +41,31 @@ def __init__(self, settings=None, loop=None):
         await asyncio.sleep(0.1)
 
     async def real_subscribe(self, channel_name):
-        try:
-            while channel_name in self._subscribers:
+        while channel_name in self._subscribers:
```
ableeb: nice

PR #653, opened 2019-08-31
Asgi: no aiohttp

vangheem:
Hi @masipcat!

I removed the aiohttp dependency.

Can you review, merge into your branch and then squash and rebase off of master?


vangheem:
ouch, I was working on an old branch version...

Let me see what I can do here.

vangheem:
argh! You already did almost all this work :)

PR #654, opened 2019-08-31
Asgi support: no aiohttp

vangheem:
For now, we can just remove the abstract base classes.

I don't need to refactor all the code necessarily.


vangheem:
Yes, we should probably add that again.

We should probably consider doing cythonize to a number of functions--measure first, see how much using cython improves it.

PR #658, opened 2019-09-01
docs(training): improve running.md

svx:
# About

This PR improves the training docs about how to run Guillotina for the first time.

# Changes

- Add link
- Adjust wording
- Remove *weasel* words
- Improve consistency by using *you* through the whole document in place from a mix of *you* and *we* 

# Rationale

## Add link

- Add link to https://www.postgresql.org
- Adjust *postgresql* to *PostgreSQL*, since that is the *official* way :)

## Wording

Replace *however* with *nevertheless*, *however* is considered to be less friendly, using *nevertheless* is a bit more supporting and friendly.

## Remove weasel words

Words like *quickly* or *potentially* are not adding a real value, but they could lead to confusion

## Improve consistency

It is always better to address the user clearly :)
The user is following the training, because of that it is better to use *you* consistent through the whole docs :) 

# Suggestion

Rearrange the order and structure to make it a bit *easier* to read through and maybe add some visual, like a picture of a browser screen at the end, when people should have it running and how that would look like :)

### DIFF
--- a/docs/source/training/running.md
+++ b/docs/source/training/running.md
@@ -1,51 +1,50 @@
 # Starting Guillotina
 
-Once you have [guillotina installed](./installation.html), you can easily run it
+Once you have [guillotina installed](./installation.html "Link to install docs"), you can run it
 with the `g` executable that it installs.
 
-However, before we begin, we'll need to run a postgresql server for Guillotina
-to use.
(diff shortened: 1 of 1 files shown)


vangheem:
Thank you!

PR #666, opened 2019-09-03
Keep imports sorted with Isort

masipcat:


masipcat:
I used a configuration that is quite similar to the current style, but I can tune it whatever you want

vangheem:
we actually have a test failure. I noticed that sometimes isort and black step on each other so sometimes you need to add some isort ignores in.

masipcat:
All done!

vangheem:
Thank you!

PR #675, opened 2019-09-18
Fix patch field delete to handle when value is None

vangheem:


vangheem:
With all these delays on travis, maybe we should try github actions...

PR #686, opened 2019-10-07
Fix iter_data mathod for DBFile(file must be DBFile)

nilbacardit26:
I found that when trying to download DBFiles using FileManager:

`AttributeError: 'DBFile' object has no attribute 'get'`

### DIFF
--- a/guillotina/tests/test_attachment.py
+++ b/guillotina/tests/test_attachment.py
@@ -5,7 +5,9 @@ from guillotina.component import get_multi_adapter
 from guillotina.interfaces import IFileManager
 from guillotina.tests import utils
 from guillotina.transactions import transaction
+from guillotina.files import FileManager
 
+import base64
 import json
 import pytest
 import random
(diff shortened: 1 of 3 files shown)


guillotina/tests/test_attachment.py
```diff
@@ -27,6 +29,33 @@
...
+            headers={"x-upload-size": str(1024 * 1024 * 10)},
+        )
+        assert status == 200
+        request = utils.get_mocked_request(db=requester.db)
+        root = await utils.get_root(request)
+        container = await root.async_get("guillotina")
+        obj = await container.async_get("foobar")
+        behavior = IAttachment(obj)
+        await behavior.load()
+        file_manager = FileManager(obj, request, behavior.file)
```
vangheem: Right here, `behavior.file` is wrong.

It should be `IAttachment['file'].bind(behavior)`

nilbacardit26: Yeap, you are right! thanks for pointing that out. I am closing this

PR #688, opened 2019-10-09
improve cache hit performance

vangheem:


guillotina/db/transaction.py
```diff
@@ -43,9 +43,10 @@ class Status:
 
 
 class cache:
-    def __init__(self, key_gen, check_state_size=False):
+    def __init__(self, key_gen, check_state_size=False, additional_keys=None):
```
ksekou: Apparently `key_gen` and `additional_keys` have the same purpose, I think we will have a more consistent API if we just use `key_gen` as a 
typing.Union[typing.Callable, typing.List[typing.Callable]].

I think we should add a docstring to describe the signature call to expect for the decorated function. That will really help new contributors like myself.

guillotina/db/transaction.py
```diff
@@ -61,6 +62,8 @@ def __call__(self, func):
                 try:
                     if not this.check_state_size or len(result["state"]) < self._cache.max_cache_record_size:
                         await self._cache.set(result, **key_args)
+                        for key_gen in this.additional_keys:
+                            await self._cache.set(result, **key_gen(result))
```
masipcat: Caching an object with one additional key, will count the cache as using `2*size_of(obj)` ?

masipcat: Therefore, reducing the effective cache size by two?

vangheem: You're right. I'll need to think about this differently.

guillotina/db/transaction.py
vangheem: This is a little bit of a difference in that `key_gen` is used to get it from the parameters, `additional_keys` sets the key/value based on the value itself.

So they are different signatures.

But yes, type hints will help and maybe I can improve the naming as well.

vangheem:
@masipcat 

> Also, right now we would only invalidate the main key:
> https://github.com/plone/guillotina/blob/76e2e20b37d142e957f0b76b20859adc5c9b010b/guillotina/contrib/cache/strategy.py#L126-L132
> 
> so we'll keep the outdated version for the additional_key.

That is actually not invalidation. That is setting the new values. However, that should also be updated.

Keys to invalidate is here:
https://github.com/plone/guillotina/blob/76e2e20b37d142e957f0b76b20859adc5c9b010b/guillotina/contrib/cache/strategy.py#L99

and here: https://github.com/plone/guillotina/blob/76e2e20b37d142e957f0b76b20859adc5c9b010b/guillotina/db/cache/base.py#L85

vangheem:
@masipcat @bloodbare can you review again? I have solved the problem where total cache size would be affected.

I also uncovered a nasty bug where we wouldn't publish keys to invalidate if you set the value. This is bad because you still need to publish the invalidation keys no matter what(because you need to invalidate in-memory data)

Please let me know how this looks...

guillotina/contrib/cache/utility.py
```diff
@@ -91,18 +91,23 @@ def get_size(self, value):
...
-            logger.info("set {} in cache".format(key))
-        except Exception:
-            logger.warning("Error setting cache value", exc_info=True)
+    async def set(self, keys, value, ttl=None):
+        if not isinstance(keys, list):
+            keys = [keys]
+        size = self.get_size(value)
+        for key in keys:
+            try:
+                size = self.get_size(value)
```
masipcat: Is this line redundant ? The size is calculated in line 97

vangheem: You're right. Fixed and added test.

vangheem:
@masipcat brought up another important point. This implementation is still flawed in that if a shared cache value with non-size 0 gets revoked, it will incorrectly calculate the size.

He pointed out that we already do that here: https://github.com/plone/guillotina/blob/master/guillotina/contrib/cache/lru.c#L381

which is dangerous!

(there is a place where we use the `cache['foo'] = 'bar'` syntax to set values which I've fixed in this PR as well.

Here are our options:

1. Stick with this approach. Cache sizing already approximate(see size calulation) and it will stay that way. The auxiliary zero-sized keys are ones that are already less used so the likelihood the cache poison us is small.

2. Go back to the original side affect of the object being counted twice. So cache sizes look larger but we know we won't use too much memory ever then and have no cache poisoning effect.

3. Fix the approach to be able to use multiple keys to point at 1 value. This requires modifying the C module(something I'm not super keen on).

4. Move to Cython... Not keen here on a point release again. Maybe we could do some speed tests on something like this to see how much trade-off there is in this.


Since maybe I'm a bit lazy, think our cache size implementation is already a bit wonky and know I've fixed a couple other critical cache related bugs in this PR, I prefer to go with `#1`.

But I can definitely be convinced otherwise.

ksekou:
I'm okay with option #1 for a short term solution, but think we should probably consider reimplementing `lru.c` in something like `Rust` for instance.
Mainly because I think it will be more maintainable, as fast as the `C` implementation and also memory safe.

vangheem:
Rust idea sounds fun :)

PR #689, opened 2019-10-11
Handle deserialization errors on bucket fields

vangheem:
Otherwise, dict value…s were getting incorrectly saved


vangheem:
@bloodbare can we consider removing this? https://github.com/plone/guillotina/blob/master/guillotina/json/deserialize_content.py#L133

It's rather dangerous...

ksekou:
Reviewing

PR #690, opened 2019-10-11
Fix connection leaks in edge-cases

masipcat:


guillotina/traversal.py
```diff
@@ -286,6 +286,7 @@ def __init__(self, resource, request, view):
                 view_result = await self.view()
                 await commit(warn=False)
             except (ConflictError, TIDConflictError):
+                await abort()
```
vangheem: The tpc handling in `transaction.py` already does some abort logic. Are we sure this is necessary here?

guillotina/db/transaction_manager.py
```diff
@@ -138,7 +138,9 @@ def lock(self):
                         raise
             except Exception:
                 # failsafe terminate to make sure connection is cleaned
-                if txn._db_conn is not None:
+                if txn._db_conn is None:
+                    raise
+                if txn._db_conn._con is None:
```
vangheem: What is this case? Connection holder doesn't actually have a connection?

How does the asyncpg pool handle this case?

Should we still call `await self._storage.terminate(txn._db_conn)` on it?

guillotina/db/storages/pg.py
```diff
@@ -782,21 +782,21 @@ def objects_table_name(self):
...
-            await shield(asyncio.wait_for(self.pool.release(con, timeout=1), 1))
+            await shield(self.pool.release(con, timeout=1))
         except (
             asyncio.CancelledError,
-            RuntimeError,
             asyncio.TimeoutError,
             asyncpg.exceptions.ConnectionDoesNotExistError,
         ):
-            pass
+            log.warning("Exception on connection close", exc_info=True)
```
vangheem: Should we also add fallback handler to terminate?

I suppose we need to be ultra careful to always release connection.

guillotina/db/storages/pg.py
```diff
@@ -782,21 +782,21 @@ def objects_table_name(self):
 
     @restart_conn_on_exception
     async def open(self):
-        conn = await self.pool.acquire(timeout=self._conn_acquire_timeout)
+        conn = await self.pool.acquire()
```
masipcat: The timeout argument "is not safe". This is implemented in the asyncpg as a `asyncio.wait_for(_acquire(), timeout=timeout)`. The awaitable in `wait_for()` runs in another task. This means that the connections is acquired from the pool in another task and returned to the "parent task". The problem here is that when the parent task is cancelled it won't get the result of the `wait_for()` and leak the connection (because the 'wait_for' task isn't cancelled). 

The issue is reported to asyncpg but isn't solved: https://github.com/MagicStack/asyncpg/pull/468

guillotina/db/storages/pg.py
```diff
@@ -782,21 +782,21 @@ def objects_table_name(self):
...
         return conn
 
     async def close(self, con):
         try:
-            await shield(asyncio.wait_for(self.pool.release(con, timeout=1), 1))
+            await shield(self.pool.release(con, timeout=1))
```
vangheem: I wonder why `wait_for` was used along with the release timeout... What does this cause that is a problem?

guillotina/db/storages/pg.py
vangheem: Are we sure we don't want to timeout here?

Otherwise, this could hang indefinitely.

guillotina/db/storages/pg.py
masipcat: I think this is redundant and... I don't trust `asyncio.wait_for()` anymore :P

guillotina/db/storages/pg.py
```diff
@@ -782,21 +782,21 @@ def objects_table_name(self):
...
         except (
             asyncio.CancelledError,
-            RuntimeError,
             asyncio.TimeoutError,
```
masipcat: asyncpg 0.19 has rewritten the RuntimeError as InterfInterfaceError

guillotina/db/transaction_manager.py
```diff
@@ -138,7 +138,9 @@ def lock(self):
...
                     raise
                 try:
                     await self._storage.terminate(txn._db_conn)
```
masipcat: This is to prevent this error:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/guillotina/db/transaction_manager.py", line 128, in _close_txn
    await self._storage.close(txn._db_conn)
  File "/usr/local/lib/python3.7/site-packages/guillotina/db/storages/pg.py", line 796, in close
    await shield(self.pool.release(con, timeout=1))
  File "/usr/local/lib/python3.7/site-packages/asyncpg/pool.py", line 645, in release
    return await asyncio.shield(ch.release(timeout), loop=self._loop)
  File "/usr/local/lib/python3.7/site-packages/asyncpg/pool.py", line 216, in release
    raise ex
  File "/usr/local/lib/python3.7/site-packages/asyncpg/pool.py", line 206, in release
    await self._con.reset(timeout=budget)
  File "/usr/local/lib/python3.7/site-packages/asyncpg/connection.py", line 1115, in reset
    await self.execute(reset_query, timeout=timeout)
  File "/usr/local/lib/python3.7/site-packages/asyncpg/connection.py", line 273, in execute
    return await self._protocol.query(query, timeout)
  File "asyncpg/protocol/protocol.pyx", line 301, in query
  File "asyncpg/protocol/protocol.pyx", line 659, in asyncpg.protocol.protocol.BaseProtocol._check_state
asyncpg.exceptions._base.InterfaceError: cannot perform operation: another operation is in progress
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/asyncpg/pool.py", line 214, in release
    self._con.terminate()
AttributeError: 'NoneType' object has no attribute 'terminate'
```

guillotina/traversal.py
masipcat: Without adding this, each ConflictError logged in sentry is one connection leaked.

guillotina/db/storages/pg.py
vangheem: :+1:

guillotina/db/storages/pg.py
vangheem: lovely

vangheem: can you leave in `RuntimeError` in case we aren't on 0.19?

guillotina/db/transaction_manager.py
vangheem: :+1:

guillotina/traversal.py
vangheem: okay, so this is fallback to make sure it is completely closed.

guillotina/db/storages/pg.py
masipcat: I'm not sure if this is a problem or not, I can put it back

guillotina/db/storages/pg.py
masipcat: This is the fallback handler: https://github.com/plone/guillotina/pull/690/files#diff-cf6270d3de166f3eb7ec0d3cce4e4ac2R140

guillotina/db/storages/pg.py
masipcat: Any idea?

guillotina/db/transaction_manager.py
masipcat: I think that this conditional is impossible to happend:
```
if txn._db_conn is None:
          raise
```
but i'm not sure about this:
```
if txn._db_conn._con is None:
    raise
```
Basically, it depends on which is the exception that raised `asyncpg.pool.release()`. I guess the connection might be already terminated or not...

guillotina/traversal.py
masipcat: Here is the traceback:
```
2019-10-11T11:52:42.996252377Z Traceback (most recent call last):
2019-10-11T11:52:42.996257215Z   File "/usr/local/lib/python3.7/site-packages/guillotina/factory/app.py", line 151, in _handle
2019-10-11T11:52:42.996261229Z     return await super()._handle(request)
2019-10-11T11:52:42.996264936Z   File "/app/api/utils/sentry.py", line 53, in sentry_app_handle
2019-10-11T11:52:42.996268793Z     return await inner()
2019-10-11T11:52:42.99627238Z   File "/app/api/utils/sentry.py", line 49, in inner
2019-10-11T11:52:42.996276471Z     reraise(*_capture_exception(hub))
2019-10-11T11:52:42.996280525Z   File "/usr/local/lib/python3.7/site-packages/sentry_sdk/_compat.py", line 57, in reraise
2019-10-11T11:52:42.996284362Z     raise value
2019-10-11T11:52:42.996287825Z   File "/app/api/utils/sentry.py", line 45, in inner
2019-10-11T11:52:42.996291375Z     response = await old_handle(self, request)
2019-10-11T11:52:42.996295131Z   File "/usr/local/lib/python3.7/site-packages/aiohttp/web_app.py", line 458, in _handle
2019-10-11T11:52:42.996298855Z     resp = await handler(request)
2019-10-11T11:52:42.996302468Z   File "/usr/local/lib/python3.7/site-packages/aiohttp/web_middlewares.py", line 119, in impl
2019-10-11T11:52:42.996306198Z     return await handler(request)
2019-10-11T11:52:42.996309825Z   File "/usr/local/lib/python3.7/site-packages/guillotina_prometheus/middleware.py", line 25, in __call__
2019-10-11T11:52:42.996325607Z     resp = await self.handler(request)
2019-10-11T11:52:42.996329334Z   File "/usr/local/lib/python3.7/site-packages/guillotina/traversal.py", line 286, in handler
2019-10-11T11:52:42.99633304Z     view_result = await self.view()
2019-10-11T11:52:42.996336637Z   File "/usr/local/lib/python3.7/site-packages/guillotina/api/service.py", line 150, in _call_original
2019-10-11T11:52:42.996340332Z     return await self.__original__()
2019-10-11T11:52:42.996343808Z   File "/app/api/services/upsert.py", line 19, in __call__
***
2019-10-11T11:52:42.996347471Z     await get_transaction().commit()
***
2019-10-11T11:52:42.996350967Z   File "/usr/local/lib/python3.7/site-packages/guillotina/db/transaction.py", line 308, in commit
2019-10-11T11:52:42.996354716Z     return await self._commit()
2019-10-11T11:52:42.996358213Z   File "/usr/local/lib/python3.7/site-packages/guillotina/db/transaction.py", line 324, in _commit
2019-10-11T11:52:42.996361855Z     await self.tpc_commit()
2019-10-11T11:52:42.996365698Z   File "/usr/local/lib/python3.7/site-packages/guillotina/db/transaction.py", line 380, in tpc_commit
2019-10-11T11:52:42.996369382Z     await self._store_object(obj, oid)
2019-10-11T11:52:42.996372906Z   File "/usr/local/lib/python3.7/site-packages/guillotina/db/transaction.py", line 366, in _store_object
2019-10-11T11:52:42.996376741Z     await self._manager._storage.store(uid, serial, writer, obj, self)
2019-10-11T11:52:42.996380556Z   File "/usr/local/lib/python3.7/site-packages/guillotina/db/storages/pg.py", line 904, in store
2019-10-11T11:52:42.996384249Z     writer,
2019-10-11T11:52:42.996388484Z guillotina.exceptions.TIDConflictError: Mismatch of tid of object being updated. This is likely caused by a cache invalidation race condition and should be an edge case. This should resolve on request retry..
```

Looking closely to the traceback I see that I'm calling (in a custom service) `get_transaction().commit()` and now I see that the logic to abort a commit is in the transaction manager. If i'd rewrite to `transaction_manager.commit(txn)` this would solve the problem. 

Today I learned that `txn.commit()` is not the same as `tm.commit(txn)` :P

guillotina/db/storages/pg.py
bloodbare: Could we shield the task and in cancelation exception cancel the inner task?

masipcat: Do you mean something like this?
```
try:
    fut = shield(self.pool.acquire(timeout=...))
    await fut
except CancelledError:
    if fut.done():
        # get connection from result and release it
    else:
        fut.cancel()
    raise
```

guillotina/traversal.py
vangheem: ouch yes, we really never want to call `.commit()` directly on a transaction object usually...

guillotina/db/transaction_manager.py
vangheem: alright, as long as it's safe. Was just wonder if it was possible to be closed by not released or something weird like that.

guillotina/db/storages/pg.py
vangheem: Doing it `conn = await self.pool.acquire(timeout=self._conn_acquire_timeout)` isn't safe because it's doing a wait for and it can get cancelled? Is that the problem?

guillotina/traversal.py
masipcat: And is there any reason to put the logic to close the connection in the tm instead of the txn itself? it looks the same to me...

vangheem: I think we could probably refactor tm out of guillotina but it would be quite a change.

Right now, tm is responsible for managing transactions and the connections associated with it.

I think it's not a great transient dependency management we put on tm to do it this way with the connection. But alas, we live with the poor decisions we've made and have to be careful about changing them later.

I would be open to it for G6!

guillotina/db/storages/pg.py
masipcat: I hope this clarifies the race condition:

f3187af5_IMG_20191011_184617

guillotina/traversal.py
masipcat: :+1:

vangheem:
What do you think about using async timeout instead? https://pypi.org/project/async_timeout/

masipcat:
I think this would work:

```python
async def open(self):
    async with timeout(self._conn_acquire_timeout):
        return await self.pool.acquire()
```

PR #695, opened 2019-10-16
Fix push invalidation data type

vangheem:
@bloodbare I don't think we have any tests in this part of our cache invalidation...

I don't use this but @masipcat said this happened with latest release.


vangheem:
This was the bug: https://github.com/plone/guillotina/pull/688/files#diff-801f19036075c76363c9d3ce31ed391bR128

PR #698, opened 2019-10-17
Add BucketDictValue.clear()

Qiwn:


guillotina/fields/annotation.py
```diff
@@ -309,6 +312,13 @@ def get_annotation_name(self, bucket_id: str) -> str:
         bucket["len"] = len(annotation["keys"])
         annotation.register()
 
+    async def clear(self, context):
```
vangheem: This will return all the annotations on the object.

What you want to do is something like this:

```
for bk in self.buckets:
  annotation_name = self.get_annotation_name(bk["id"])
  await annotations_container.async_del(annotation_name)
```

guillotina/tests/test_annotation.py
```diff
@@ -0,0 +1,43 @@
+from guillotina.content import create_content_in_container
```
vangheem: I think we already have a `test_annotations.py` file -- maybe just add it here?

Or, the other bucket field tests are in `test_serialize.py`

vangheem:
Fixed flake8 issue and rebased.

PR #699, opened 2019-10-17
fix error closing the cache on some objects

vangheem:
This also exposed an error with this implementation because we were doing `self.get_key(container=obj.__parent__, id=obj.__name__)` when `obj.__parent__` was `None`. So we were setting invalid cache values.

```
WARNING:guillotina:Error closing connection
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/guillotina/contrib/cache/strategy.py", line 125, in close
    await self.fill_cache()
  File "/usr/local/lib/python3.7/site-packages/guillotina/contrib/cache/strategy.py", line 148, in fill_cache
    val["parent_id"] = obj.__parent__.__uuid__
AttributeError: 'NoneType' object has no attribute '__uuid__'
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,8 @@ CHANGELOG
 5.0.24 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Fix error closing cache with some objects
+  [vangheem]
 
 
 5.0.23 (2019-10-17)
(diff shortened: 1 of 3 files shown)


vangheem:
@masipcat I think it's only possible on the root object. I saw it running a db migration which writes to the root object to track the version.

PR #717, opened 2019-10-30
Added deserializer for IUnionField

masipcat:


masipcat:
I'm doing a content type that has a field like this:

```python
class IObjectA(Interface):
    foo = schema.Text(required=False)

class IObjectB(Interface):
    bar = schema.Text(required=False)

class IMainObject(Interface):
    union = schema.UnionField(Object(IObjectA), Object(IObjectB))
```

The problem is that IObject deserializer never fails if receives a dictionary that doesn't match any field, it just deserializes the value to `{}` (that it's why I added this: https://github.com/plone/guillotina/pull/717/commits/687549180bde5ffc6c03fac732f1db8e58ea7afc#diff-b2a845fe39a48b7b8c8bbeb584780abaR153).

I was thinking to modify the IObject deserializer to:
```python
@configure.value_deserializer(IObject)
def object_converter(field, value, context=None):
    if not isinstance(value, dict):
        raise ValueDeserializationError(field, value, "Not an object")
    result = {}
    for key, val in value.items():
        if key in field.schema:
            f = field.schema[key]
            if val is not None:
                result[key] = get_adapter(f, IJSONToValue, args=[val, context])
            else:
                result[key] = None
+       else:
+           raise ValueDeserializationError(field, value, f"Field {key} not found")
    return result
```
But I think this could potentially break other things...

masipcat:
Maybe we could add the argument `strict` to Object field to support both behaviors

vangheem:
I'm okay with added `strict` keyword.

FWIW, json schema field is much faster than the object field and might provide the flexibility you need.

masipcat:
> FWIW, json schema field is much faster than the object field and might provide the flexibility you need.

I'll give it a try when I need to create another content type ;-)

PR #718, opened 2019-11-01
Provide new `@field.validator` to validate field values against bound fields

vangheem:
Also, fix invariant implementation.

### DIFF
--- a/docs/source/developer/fields.md
+++ b/docs/source/developer/fields.md
@@ -388,7 +388,7 @@ from zope.interface import Interface
 from guillotina import schema
 
 class IMySchema(Interface):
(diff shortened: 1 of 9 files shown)


docs/source/developer/fields.md
```diff
@@ -370,3 +370,63 @@ Delete:
...
+validated for the field.
+
+The most simple type of constraint is to use a lambda function:
+
+```python
+from zope.interface import Interface
+from guillotina import schema
+
+class IMySchema(Interface):
+    field = schame.Text(constraint=lambda val: val != 'foobar')
```
ableeb: should be `schema`

PR #720, opened 2019-11-02
Discordance between the attributes defined in __slots__ and the actual attributes

masipcat:


guillotina/db/orm/base.py
```diff
@@ -71,16 +71,15 @@ def __repr__(self):
...
-        "__of",
-        "__name",
-        "__annotations",
-        "__immutable_cache",
-        "__new_marker",
-        "__txn",
-        "__uuid",
-        "__serial",
-        "__volatile",
+        "__parent__",
```
masipcat: Or should this be `_BaseObject__parent`?

masipcat:
> I thought this was how slots worked.
>
> Anyways, __parent__, __of__, etc are all properties. The actual data is stored in the slot though.

Oh, my bad, I didn't know that slots like `__attr` was actually stored in `inst._TheClass__attr`.

> Do you see a problem that you are trying to fix?

No, I haven't experienced any problem.
Sorry!

vangheem:
No problem! @masipcat thank you for sanity checking it.

I'm not saying there's no chance it's wrong. We could have done it wrong.

I think this is how slots are implemented as a base class though.

vangheem:
and I know it's a bit wonky...

PR #723, opened 2019-11-06
Add extra_headers parameter into FileManager.prepare_download()

Qiwn:
I added this change because I can't add more headers into a prepared StreamResponse


vangheem:
5.1.4 released.

PR #724, opened 2019-11-06
Add db in task vars when getting container in tests

lferran:


bloodbare:
Should not transaction set the db on task vars inside the context?

vangheem:
@bloodbare are we sure that a txn manager should be setting task vars of a db?

We get the txn manager from the db so the relationship is a bit reversed.

vangheem:
@bloodbare also, are we sure we'd want to change this behavior in a PR just to help test setup?

bloodbare:
Sorry about it, seeing a context manager needing to set a context var was disruptive, I would expect that opening a transaction on a DB makes sure that DB is on a txn var.  You mean that tests are broken? I don't see it on Travis.

Issue #722, opened 2019-11-04
Python 3.8 and pickles


masipcat:
I started testing Guillotina on Python 3.8. Python 3.8 introduced pickle protocol v5. Since guillotina is configured to use `pickle.HIGHEST_PROTOCOL`, all the objects serialized with py3.8 aren't compatible with py3.7 so is not possible to run instances of guillotina with different python versions side by side.

Thoughts about this:
1. Maybe using HIGHEST_PROTOCOL its a bit dangerous and we could choose the protocol version based on the python versions we support (for instance, G5 supports python 3.7 and 3.8, so we could choose to use protocol v4). We can upgrade the protocol version on G6 or G7.
2. Allow to configure protocol version through configuration (so the user can choose to use HIGHTEST_PROTOCOL if he wants)

cc/ @vangheem @bloodbare

vangheem:
ouch, yes, I agree.

Let's by default use v4 and then make it configurable.

We'll also need to be careful to consider this when(if) we deprecate a protocol.

PR #725, opened 2019-11-06
Added "pickle_protocol" to app_settings

masipcat:
Close https://github.com/plone/guillotina/issues/722

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,8 @@ CHANGELOG
 5.1.4 (unreleased)
 ------------------
 
-- Nothing changed yet.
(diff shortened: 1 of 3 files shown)


bloodbare:
4 and 3 are 100% comptible?

vangheem:
@bloodbare version 4 was added in py 3.4 and is what we've been using.

The problem is with 3.8 which introduces version 5.

@masipcat now that I'm thinking about it, I wonder if defaulting to highest protocol would be better.

It's an unusually use-case to want to support multiple python versions and we should probably default to 5 when possible since it has performance improvements.

masipcat:
Merge? :P

PR #728, opened 2019-11-11
Make sure to use `txn.lock` when using pg connection

vangheem:


vangheem:
@masipcat Imagine doing something that uses pg in a gather or if you spawn an async task that uses pg.

It's really important to make sure to never simultaneously use the connection with something else.

PR #738, opened 2019-11-21
Debug subscribers

vangheem:
@bloodbare thoughts on this approach?

You would be responsible for the logging configuration to do something with this data.

### DIFF
--- a/guillotina/component/globalregistry.py
+++ b/guillotina/component/globalregistry.py
@@ -51,7 +51,7 @@ class GuillotinaAdapterLookup(AdapterLookup):
         return result
 
 
-class DebugGuillotinaAdapterLookup(AdapterLookup):
+class DebugGuillotinaAdapterLookup(GuillotinaAdapterLookup):  # pragma: no cover
     @profilable
     async def asubscribers(self, objects, provided):
         from guillotina.utils import get_current_request, get_authenticated_user_id, get_dotted_name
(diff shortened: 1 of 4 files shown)


bloodbare:
Whould be great some documentation and instructions how to use it (besides changelog)

guillotina/component/globalregistry.py
```diff
@@ -46,13 +51,73 @@ def subscribers(self, objects, provided):
...
+
+
 class GuillotinaAdapterRegistry(AdapterRegistry):
     """
     Customized adapter registry for async
     """
 
     _delegated = AdapterRegistry._delegated + ("asubscribers",)  # type: ignore
-    LookupClass = GuillotinaAdapterLookup
+    if os.environ.get("DEBUG_SUBSCRIBERS") in ("1", "true", "t") or True:
```
masipcat: Is the `or True` correct?

guillotina/component/globalregistry.py
```diff
@@ -46,13 +51,73 @@ def subscribers(self, objects, provided):
...
+        except RequestNotFound:
+            request = None
+        try:
+            url = request.url.human_repr()
+        except AttributeError:
+            # older version of aiohttp
+            url = ""
+        info = {
+            "url": url,
+            "account": getattr(task_vars.container.get(), "id", None),
```
bloodbare: should be container

guillotina/component/globalregistry.py
vangheem: Nope! Fixed :)

guillotina/component/globalregistry.py
```diff
@@ -46,13 +51,73 @@ def subscribers(self, objects, provided):
...
+
+
 class GuillotinaAdapterRegistry(AdapterRegistry):
     """
     Customized adapter registry for async
     """
 
     _delegated = AdapterRegistry._delegated + ("asubscribers",)  # type: ignore
-    LookupClass = GuillotinaAdapterLookup
+    if os.environ.get("GDEBUG_SUBSCRIBERS") in ("1", "true", "t"):
```
bloodbare: maybe better:

`os.environ.get("GDEBUG_SUBSCRIBERS", "").lower() in ("1", "true", "t")`

vangheem: thanks, updated!

PR #739, opened 2019-11-21
refactor read conn to exclusively be for operations and not initialization

vangheem:


masipcat:
If I'm not getting it wrong, the read_conn is only used during initialization (in `storage._initialize_tid_statements()` and `storage.create()`), right? In that case, whats the purpose of keeping this connection open?

vangheem:
It's used to get the current tid of a transaction and check conflicts as well.

vangheem:
@bloodbare it's ready for final review

PR #741, opened 2019-11-25
Make sure to reset registry task var when setting up new container

vangheem:


guillotina/contentapi.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, db, user=RootUser("root")):
         await self.abort()
 
     async def use_container(self, container: IContainer):
+        task_vars.registry.set(None)
```
lferran: is the `registry` task var set on `get_registry(container)`?

lferran: ah yeah! I see it now :+1:

vangheem: Yes, it is! https://github.com/plone/guillotina/blob/master/guillotina/utils/misc.py#L381

guillotina/commands/run.py
```diff
@@ -41,6 +42,10 @@ def get_parser(self):
             return
         sig = inspect.signature(module.run)
         if "container" in sig.parameters:
```
bloodbare: I don’t see this related with the issue

PR #744, opened 2019-11-27
Fix bug: anonymous user should not get guillotina.Authenticated

lferran:


bloodbare:
I agree that may be confussing to understand that role guillotina.Authenticated doesn't mean that is a valid user. To make this less confussing we could remove the `guillotina.Authenticated = 1`. Then on the GuillotinaUser init add `self._roles["guillotina.Authenticated"] = Allow`

lferran:
ah ok!! @bloodbare thanks for pointing that out. Yes, it was confusing to me. 

Apart from that, what default permission should I use if I want an endpoint to be accessible to all roles except those users that do not have a valid JTW token present?

guillotina/auth/utils.py
```diff
@@ -31,11 +30,6 @@
 
 
 def set_authenticated_user(user):
-    if user is not None:
```
vangheem: Instead of removing this and hardcoding `guillotina.Authenticated` into the anon user object, can we just make the `roles` not editable on the anon user object?

guillotina/auth/users.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, user_id="guillotina", groups=None, roles=None, permissions=No
         self.id = user_id
         self._groups = groups or []
         self._roles = roles or {}
+        self._roles["guillotina.Authenticated"] = Allow
```
vangheem: I don't think this is safe. `_roles` might be overridden. Adding the `guillotina.Authenticated` role as part of the auth process is on purpose.

vangheem:
@bloodbare I don't think we should hardcode `guillotina.Authenticated` into the roles of a user object. It should be part of the auth process. It might have unintended consequences if we hard code it on the user object.

guillotina/auth/utils.py
lferran: done

PR #745, opened 2019-11-28
Be able to iterate bucket dict keys, values and items

lferran:


guillotina/fields/annotation.py
```diff
@@ -326,6 +329,35 @@ def __len__(self):
...
+        async for annotation in self._iter_annotations(context):
+            for key in annotation.data["keys"]:
+                yield key
+
+    async def values(self, context) -> AsyncIterator[Any]:
+        async for annotation in self._iter_annotations(context):
+            for value in annotation.data["values"]:
+                yield value
+
+    async def items(self, context) -> AsyncIterator[Tuple[str, Any]]:
```
bloodbare: Can we provide the same interface as BucketListValue?

guillotina/fields/annotation.py
```diff
@@ -326,6 +329,35 @@ def __len__(self):
             total += bucket.get("len", 0)
         return total
 
+    async def _iter_annotations(self, context) -> AsyncIterator[AnnotationData]:
```
bloodbare: should be iter_buckets

vangheem:
5.1.21 released

PR #747, opened 2019-12-02
Fix security bug on duplicate

lferran:
By setting `check_permission: False` on the payload, caller was able to create content on destination even if he/she didn't have access to.

### DIFF
--- a/guillotina/content.py
+++ b/guillotina/content.py
@@ -700,6 +700,7 @@ async def duplicate(
         id=new_id,
         creators=context.creators,
         contributors=context.contributors,
+        check_security=check_permission,
     )
 
     for key in context.__dict__.keys():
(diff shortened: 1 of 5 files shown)


guillotina/content.py
```diff
@@ -700,6 +700,7 @@ def get_all_behavior_interfaces(content) -> list:
         id=new_id,
         creators=context.creators,
         contributors=context.contributors,
+        check_security=check_permission,
```
lferran: Do you guys think that's ok? IMHO yes, but I'd appreciate your confirmation

masipcat:
Btw, the same error exist in endpoint move

lferran:
oh ok @masipcat let me fix too!

guillotina/content.py
vangheem: Yes, this is good.

lferran:
unrelated swagger test failing :thinking: so weird

vangheem:
5.1.22 released!

PR #751, opened 2019-12-06
Added async property `Request.body_exists` and fixed fixture 'guillotina'

masipcat:


guillotina/tests/fixtures.py
```diff
@@ -499,10 +499,9 @@ def mock_txn():
 
 
 @pytest.fixture(scope="function")
-def guillotina(app_client):
+async def guillotina(app_client):
     _, client = app_client
-    requester = GuillotinaDBAsgiRequester(client)
-    yield requester
+    return GuillotinaDBAsgiRequester(client)
```
masipcat: This change + https://github.com/plone/guillotina_elasticsearch/compare/3.3.x...masipcat:3.3.x-g6 makes guillotina_elasticsearch compatible with G6!

masipcat:
> Why we need this function ? 

The idea is to implement https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.BaseRequest.body_exists to be more backwards compatible

> Will it be able to read the body after body_exist call ? Maybe we should store on a buffer on body_exist

Yes, `request.read()` consumes the stream and stores the entire body in `request._read_bytes` (https://github.com/plone/guillotina/pull/751/files#diff-cc026d1fe7a5d0392e22138ef60b2f10R504-R530)

PR #754, opened 2019-12-09
Tags from dublincore should be stored on context, any information tha…

bloodbare:
…t needs to be indexed on PG Catalog implementation needs to be on the main object otherwise PATCH does not update it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@ cov_html
 # files
 *.log
 *.swp
+.DS_Store
 
 # excludes
 !.coveragerc
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,9 @@ CHANGELOG
 6.0.0 (unreleased)
 ------------------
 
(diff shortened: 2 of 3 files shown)


vangheem:
This is dangerous. All existing content that had tags has lost its data now..

We should have a migration step.

Also, I'm not quite sure why this should matter for indexing. The indexer should load the behavior data. We use the tags index.

bloodbare:
I'm aware of it, I'm ok to have as a breaking change for G6.

AnnotationBehavior data can not be indexed right now because its indexed at saving the pickle that gets the json data of the pickle object that is saving. In case of annotation data it does not save to the resource object with metadata.

vangheem:
That's not a breaking change though, that's data loss.

Can we have a migration handler?

Better yet would be a b/w compatible field that will look at the annotation data as a backup until the data is moved.

bloodbare:
Yups no problem.

vangheem:
We probably need a doc like we had in g5 for all the breaking changes too! \cc @masipcat

PR #757, opened 2019-12-09
WIP: Migrating tags

bloodbare:
Needs to suport txn chunks

### DIFF
--- a/guillotina/db/walking.py
+++ b/guillotina/db/walking.py
@@ -1,7 +1,7 @@
-from guillotina.interfaces import IFolder
-from guillotina.interfaces import IResource
-from guillotina.interfaces import IIteratorResources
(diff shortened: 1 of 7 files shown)


guillotina/db/walking.py
```diff
@@ -0,0 +1,49 @@
...
+
+@adapter(for_=(IFolder), provides=IIteratorResources)
+class WalkingIterator(object):
+    def __init__(self, context):
+        self.context = context
+        self.total = 0
+        self.count = 0
+
+    async def deep_walk(self, actual):
+        self.total += await actual.async_len()
```
vangheem: I would recommend against using `async_len`. PG is always very slow to get total counts.

bloodbare:
No activity for a long time we close

PR #756, opened 2019-12-09
conflict on @move must return http409 error

inakip:

PR #763, opened 2019-12-10
conflict on @move must return http409 error

inakip:
backport of https://github.com/plone/guillotina/pull/756


vangheem:
@inakip looks like a few tests need to be updated.

Why didn't these fail in master?

inakip:
> @inakip looks like a few tests need to be updated.
> 
> Why didn't these fail in master?

weird, I run all tests in local before pushing and all passed
looking...

PR #764, opened 2019-12-10
Changes to middleware logic

masipcat:


docs/source/installation/configuration.md
```diff
@@ -161,10 +161,17 @@ load_utilities:
 
 ## Middleware
 
-`guillotina` is built on `asgi` which provides support for middleware.
+`guillotina` support two types of middlewares:
+
+1. ASGI middlewares: run before the guillotina asgi application and can be used to
+  catch errors or to record timing metrics
```
masipcat: > can be used to catch errors

Right now this is not true because the traversal captures all unhandled exceptions and generates the error response. Ideally, the first ASGI middleware should be the ErrorMiddleware that generates the error response, like in Starlette: https://github.com/encode/starlette/blob/7f8cd041734be3b290cdb178e99c37e1d5a19b41/docs/middleware.md

Agree?

masipcat: Also, I think that guillotina_prometheus should be rewritten as a ASGI middleware because now it runs after the traversal so the response_time/latency isn't accurate (doesn't include the time to run traversal).

jordic: Cool! That's the reason why our api service it's so fast :)

masipcat:
@vangheem @bloodbare PR ready for review

docs/source/installation/configuration.md
vangheem: A error handling middleware seems like a nice simple way to provide an example middleware!

docs/source/installation/configuration.md
```diff
@@ -161,14 +161,30 @@ load_utilities:
...
+2. Guillotina (aiohttp-like) middlewares: run after the router and before executing the view
+
 You can provide an array of dotted names to use for your application.
 
 ```yaml
+asgi_middlewares:
```
vangheem: Should we document that the order defined here matters?

For example, if we provide error handling middleware, it needs to always be last.

Also, have you thought about how to pass configuration to the middleware?

docs/source/installation/configuration.md
masipcat: I'll write it ;) The thing is that we need to decouple this logic from the traversal to middlewares. I'd say that the error and transaction logic should be in separate middlewares.

```
(incoming request) 
ErrorASGIMiddleware
[ User-provided-ASGI-middlewares ... ]
TransactionASGIMiddleware
TraversalRouterMiddleware (creates Request obj)
[ User-provided-middlewares ... ]
View execution
```

Makes sense?

docs/source/installation/configuration.md
masipcat: > For example, if we provide error handling middleware, it needs to always be last.

I'd say the ErrorMiddleware will be added dynamically

> Also, have you thought about how to pass configuration to the middleware?

The middleware can read the `app_settings` directly.

masipcat:
> Can we define pre and post middlewares? 

If you mean post view execution, no. We can define ASGI middlewares before "Guillotina", in other words, before the router/traversal. These will sit between the ASGI server and the ASGI framework. 
Then we have the same behaviors we already have, that sit between the router/traversal and the view to execute (like the aiohttp middlewares).

> Logging and error handling should be done after handling request.

In that case, the ASGI middlewares will never see a python exception therefore we can't use middlewares like Sentry.

docs/source/installation/configuration.md
vangheem: I think that should work.

I'm not sure a `TraversalRouterMiddleware` is necessary though?

masipcat: `TraversalRouterMiddleware` is the `AsgiApp` in `guillotina/asgi.py`, so this would be:

```
(incoming request) 
ErrorASGIMiddleware
[ User-provided-ASGI-middlewares ... ]
TransactionASGIMiddleware
GuillotinaAsgiApp (creates Request obj)
[ User-provided-middlewares ... ]
View execution
```

To be sure, I understand that you agree on decoupling the error and transaction logic outside of the Router/Traversal/AsgiApp , right?

vangheem:
maybe we should allow flexibility so you could insert a middleware between the error one and do custom error handling?

Also, as for pre/post... usually the way you implement middleware should be done where pre/post is part of the middleware implementation.

dummy psuedo code following:
```
async def my_middleware(app, req):
  # pre code
  resp = await app(req)
  # post code
  return resp
```

docs/source/installation/configuration.md
vangheem: Yes, error and transaction make sense.

Maybe I just need to see an implementation to understand how I'll like it.

masipcat:
@vangheem @bloodbare Ready for review!

**Changes:**
* I managed to add support for asgi middlewares and aiohttp (can be mixed). 
* Added `ErrorsMiddleware` that catch all errors and renders the json response. Is always the first middleware
* Moved the `traversal` and `resolve_route` to `TraversalRouter`
* Moved the view execution to `ServiceExecutor` (so we can run other middlewares between the traversal and the view, like in G5)

So, now we can add middlewares like Sentry and Prometheus before Traversal.

docs/source/installation/configuration.md
masipcat: Done!

masipcat:
> Logging and error handling should be done after handling request.

@bloodbare Regarding your comment, now the request is instantiated before executing any request

guillotina/middlewares/service_executor.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class ServiceExecutor:
+    """
+    Its always the last middleware and is responsible to call the actual
+    guillotina service (the view resolved during the traversal)
+    """
+
+    async def __call__(self, scope, receive, send):
+        service_handler = task_vars.service.get()
```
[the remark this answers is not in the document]
masipcat: I prefer using the scope dictionary, so we can use generic asgi middlewares that use this signature. I'll change it

guillotina/middlewares/errors.py
```diff
@@ -0,0 +1,65 @@
...
+
+class ErrorsMiddleware:
+    def __init__(self, app):
+        self.next_app = app
+
+    async def __call__(self, scope, receive, send):
+        headers_sent = False
+
+        async def _send(msg):
+            nonlocal headers_sent
```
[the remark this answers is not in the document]
masipcat: Yes.. we need to do this to know if the service (or another middleware) already sent the headers

guillotina/asgi.py
```diff
@@ -83,6 +88,36 @@ def __call__(self, scope, receive=None, send=None):
...
+        # We instantiate middlewares in reverse order
+        for middleware in reversed(user_middlewares):
+            args = inspect.getargspec(middleware).args
+            if args[-1] == "handler":
+                if "app" in args:
+                    middleware = aiohttpHandler2asgi(middleware)
+                else:
+                    middleware = aiohttp2asgi(middleware)
+
+            last_middleware = middleware(last_middleware)
```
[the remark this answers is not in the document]
masipcat: But before overwriting last_middleware we pass it to the previous middleware constructor

masipcat:
The thing is:

* aiohttp/G5 runs the middlewares **after** the traversal/router and **before** the service.
* G6 runs the middlewares before or after the traversal/router. I've done it that way to maintain compatibility with G5.

However, I don't see the benefit from running middlewares after the TraversalRouter. This logic can be implemented in a custom service class or in a custom decorator.

So, if you agree, I can remove the ServiceExecutor and drop support for middlewares after the router. This would reduce the changes and complexity.

guillotina/asgi.py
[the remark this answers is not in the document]
vangheem: ah right, sorry!

masipcat:
**Final changes:**

Now we only support ASGI middlewares that run before the Traversal/Router and before instantiating the request.

This would require updating guillotina_prometheus (do we have more middlewares?). I'll take care.

masipcat:
In a future PR I'll try to move the CORS logic and Debug headers into a middleware to reduce the code in the traversal. Is it ok?

vangheem:
I prefer it be a wrapper. That's how I understand middleware is supposed to work. 

Another thought I had is that maybe we could make sure our middleware is compatible with starlette middleware.

bloodbare:
> I prefer it be a wrapper. That's how I understand middleware is supposed to work.
> 
> Another thought I had is that maybe we could make sure our middleware is compatible with starlette middleware.

Yups,  I mean defining that some functionality of G can be exported as middlewares on wrappers (traversals, .... ) I feel its going to be overengineering.

vangheem:
ahh, gotcha. I had the same worry then kind of. @masipcat how did you feel about it after moving it? Did it seem clean or a bit odd to move those bits into middelware?

I can try to spend some more time thinking about it, reviewing and seeing how I feel about it.

jordic:
From my humil point of view, thought we should go on g6 with what we have.. later when we use it for a time perhaps we can evolve a bit more (like g7?)

masipcat:
Thank you for reviewing!

---

@bloodbare I'm ok with not moving the X-Debug logic and other internals  to a middleware :) 

----

> Another thought I had is that maybe we could make sure our middleware is compatible with starlette middleware.

@vangheem this would be great, but I see some problems:
1. In Starlette the middlewares use their Request/Response objects. 
3. In guillotina, the request is created after all middlewares ran. I'd like to implement the Request as a proxy to the asgi `scope` and create the request before running the middlewares, but it's a bit tricky.
4. Starlette uses https://www.starlette.io/middleware/#basehttpmiddleware in some middlewares (it's a abstraction that allows you to write middlewares easily, but it caused some problems in our projects -- each `call_next()` run the next middleware in another asyncio task... Maybe @jordic rembemers better the issues we faced)

> @masipcat how did you feel about it after moving it? Did it seem clean or a bit odd to move those bits into middelware?

Good, it's not the first asgi middleware I write and the ErrorMiddleware is the perfect use-case for a middleware. Also, I feel it happens too much in the Traversal and IMHO we should move the "http related logic outside the "guillotina app"/traversal. Like CORS :P

----

Finally, I 100% agree with @jordic . Now we have the minimal implementation to support generic asgi middlewares. 

Maybe we should discuss where we want to go. Do we want to be a http framework that is compatible with asgi servers? Do we want to build guillotina on top of starlette (like FastAPI)? or maybe we only want to use their Request/Response to have compatibility with their Middlewares.

vangheem:
Okay, I agree. I'm going to approve.

PR #768, opened 2019-12-11
Move token expiration to a app_setting

jordic:


vangheem:
Looks like black issue!

PR #769, opened 2019-12-12
Provide an endpoint to list global roles.

jordic:
With this, from the api, a user with permissions to manage users,
can assign global roles to users and groups


jordic:
Still no tests, but, just to check is what we need to be able to discover the available roles.
Just in case, Are local roles needed on any part? For example to assign it to a container or item?
Should this function also be available from an UI?

PR #772, opened 2019-12-13
@duplicate: added option to reset acl

inakip:


guillotina/content.py
```diff
@@ -694,26 +696,36 @@ def get_all_behavior_interfaces(content) -> list:
...
         if key.startswith("__") or key.startswith("_BaseObject"):
             continue
-        if key in ("id",):
+        if key in ("id", "creators", "contributors"):
             continue
         new_obj.__dict__[key] = context.__dict__[key]
+
     new_obj.__acl__ = context.__acl__
+    if reset_acl:
+        prinrole = IPrincipalRoleManager(context)
```
vangheem: Should use the IGetOwner:

```
get_owner = get_utility(IGetOwner)
        roleperm = IPrincipalRoleManager(new_obj)
        owner = await get_owner(new_obj, user)
        if owner is not None:
            roleperm.assign_role_to_principal("guillotina.Owner", owner)
```

guillotina/content.py
```diff
@@ -694,26 +696,36 @@ def get_all_behavior_interfaces(content) -> list:
...
-
     for key in context.__dict__.keys():
         if key.startswith("__") or key.startswith("_BaseObject"):
             continue
-        if key in ("id",):
+        if key in ("id", "creators", "contributors"):
             continue
         new_obj.__dict__[key] = context.__dict__[key]
+
     new_obj.__acl__ = context.__acl__
```
vangheem: Should be in else?

guillotina/content.py
```diff
@@ -694,26 +696,36 @@ def get_all_behavior_interfaces(content) -> list:
...
+    if reset_acl:
+        prinrole = IPrincipalRoleManager(context)
+        prinrole.map._clear()
```
vangheem: Should do `new_obj.__acl__ = None` instead!

guillotina/content.py
inakip: ok, thanks

guillotina/content.py
inakip: ok, thanks

vangheem:
@bloodbare @masipcat Are you okay with this change? Can we merge?

@inakip please also make change to 5.x!

Issue #774, opened 2019-12-15
User and Group from dbUsers should be restricted to UserFolder/GroupFolder


jordic:
Doesn't make any sense that and admin could add a user content type everywhere.. 

https://github.com/plone/guillotina/blob/master/guillotina/contrib/dbusers/content/users.py#L52

jordic:
@lferran we need to add something here, no?

lferran:
@jordic I think this issue can be closed. I already fixed in this PR: https://github.com/plone/guillotina/pull/800/files#diff-35da10df07a75358094b17645171e9b0R195

PR #777, opened 2019-12-16
Restrict users and groups only added under manager folders

lferran:
- provide way to restrict the type of the parent
- restrict parent type for users and groups

fixes: https://github.com/plone/guillotina/issues/774


guillotina/constraintypes.py
```diff
@@ -30,6 +31,28 @@ def get_allowed_types(self) -> Optional[list]:
...
 
 
+@configure.adapter(for_=Interface, provides=IConstrainParentTypes)
+class FTIConstrainAllowedParentTypes:
+    def __init__(self, context: IResource) -> None:
+        self.context = context
+
+    def is_type_allowed(self, type_id: str) -> bool:
+        if type_id in app_settings["container_types"]:
+            return False
```
lferran: not sure if this is needed in this case. can you confirm?

lferran:
good point! thanks @bloodbare

lferran:
@bloodbare is there already some functionality in place to define a type of content not globally addable?

bloodbare:
At guillotina_cms there is an implementation (https://github.com/plone/guillotina_cms/blob/master/guillotina_cms/content/contraintypes.py) that allows to define globally disabled types and assign constraints on the local object. This aproach is not using the same Interface as Guillotina as it may happen that you want that internally create_content_in_container can overwrite the restriction. For example, if you want to disable UsersFolders globally but you want to create one on addon install function you need to skip the restriction. So the steps should be to:
- Define at create_content_in_container a flag to skip this check
- Move internal code to make sure that all code calls (minus content POST) uses that flag
- Move gcms aproach of global/local constraints to guillotina

I suggest you work on a new branch as its far related to this one.

lferran:
@bloodbare oh ok! I will look at it

I've been working on an alternative solution while you were replying.

Instead of defining globally enabled/disabled types in app_settings, we could just define it in the interface factory.

what do you think about it? Isn't it more self contained?

bloodbare:
You can define on the interface factory if its going to be globaly allowed, and you can populate app_settings on factory registration

jordic:
Thinkin a bit on the dev api, perhaps makes sense something like...
```python
@configure.contenttype(
    type_name="Conversation",
    schema=IConversation,
    behaviors=["guillotina.behaviors.dublincore.IDublinCore"],
    allowed_types=['Message'],
    addable_on=(IFolder, IXXX,)
)
...

**Comment by jordic** [comment]
@vangheem What needs to be done to be able to add the parameter `addable_on` on the contenttype directive?

**Comment by lferran** [comment]
@jordic that was my initial solution, but we decided that it seemed like duplicate functionality.

Instead, @bloodbare suggested that we could define a boolean attribute in the `contentype` directive like `globally_addable=False`

Does that solution sound good for everyone?

> At guillotina_cms there is an implementation ([plone/guillotina_cms:guillotina_cms/content/contraintypes.py@`master`](https://github.com/plone/guillotina_cms/blob/master/guillotina_cms/content/contraintypes.py) [](/plone/guillotina_cms/blob/HEAD@{2019-12-16T11:17:50Z}/guillotina_cms/content/contraintypes.py)) that allows to define globally disabled types and assign constraints on the local object. This aproach is not using the same Interface as Guillotina as it may happen that you want that internally create_content_in_container can overwrite the restriction. For example, if you want to disable UsersFolders globally but you want to create one on addon install function you need to skip the restriction. So the steps should be to:
> 
> * Define at create_content_in_container a flag to skip this check
> * Move internal code to make sure that all code calls (minus content POST) uses that flag
> * Move gcms aproach of global/local constraints to guillotina
> 
> I suggest you work on a new branch as its far related to this one.

I will give it a try

**Comment by jordic** [comment]
It's not clear for me how this solves the problem, seems more like a "fix" to hide some types, but, how I can restrict a type to be addable only on some Interface.
We can declare which types can be added on something.. but seems like a clean Dev interface to be able to declare where a type can be added.

**Comment by lferran** [comment]
@jordic yes it solves the problem. You would have to define:

```python
@configure.contenttype(
    type_name="User",
    ...
    globally_addable=False,
)
```
And then have it allowed on the user manager folder:

```python
@configure.contenttype(
    type_name="UserManager",
    ...
    allowed_types=["User"],
    globally_addable=False,
)
```
Notice that `UserManager` is itself not globally addable. That's why we need to be able to change `create_content_in_container` to skip that check when creating the `UserManager` folder on install.

I am not sure what would be a better dev interface though... I let you guys decide on that :)

jordic:
Sounds good to me!
It was not clear for me the second part.
For the second, thought this should be just an API constraint.. (the user folder is installed throught the app install).
We also need to fix the @addable-types.
Thought, as a compat layer, all types are global.

lferran:
Closing this PR in favour of https://github.com/plone/guillotina/pull/800

Issue #779, opened 2019-12-17
Deploying G6 at K8s fails due to HTTP 2 keep alive problem 


bloodbare:
More information:

https://github.com/encode/uvicorn/issues/344

vangheem:
dang, I saw this sort of thing with ambassador/envoy as well :(

jordic:
Perhaps, this doesn't happens to us because we are using caddy (go based). Is envoy perhaps based on the same http lib than nginx?

PR #780, opened 2019-12-17
Fixing #779 by settings http protocol and allowing to set keep alive …

bloodbare:
…timeout

### DIFF
--- a/guillotina/_settings.py
+++ b/guillotina/_settings.py
@@ -91,7 +91,11 @@ app_settings: Dict[str, Any] = {
     "search_parser": "default",
     "object_reader": "guillotina.db.reader.reader",
     "thread_pool_workers": 32,
-    "keep_alive": 5,
-    "http_protocol": "h11"
+    "server_settings": {
+        "uvicorn": {
+            "timeout_keep_alive": 5,
+            "http": "h11"
+        }
+    }
 }
 default_settings = copy.deepcopy(app_settings)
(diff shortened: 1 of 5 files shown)


guillotina/_settings.py
```diff
@@ -91,5 +91,7 @@
     "search_parser": "default",
     "object_reader": "guillotina.db.reader.reader",
     "thread_pool_workers": 32,
+    "keep_alive": 5,
```
vangheem: 5 is very low and I'm not sure it'll actually help the situation if this is the same problem that I've seen.

fwiw, nginx defaults to 75sec keep alive I believe.

guillotina/_settings.py
```diff
@@ -91,5 +91,7 @@
...
+    "http_protocol": "h11"
```
vangheem: it's a shame we end up with h11 here

masipcat:
I'm not sure if we want to expose all these options in the cli / app_settings in the top level. I'd suggest something like:
```python
app_settings = {
    "server_settings": {
         "uvicorn": { ... }
     }
}
```
that is mapped to the uvicorn Config

guillotina/_settings.py
bloodbare: default keep alive timeout at uvicorn is 5 (https://github.com/encode/uvicorn/blob/master/uvicorn/config.py#L134)

guillotina/_settings.py
bloodbare: Its temporal until encode/uvicorn#344 is fixed

PR #788, opened 2019-12-18
Replace pytest-asyncio with alt-pytest-asyncio

masipcat:
What do you think about using this library https://github.com/delfick/alt-pytest-asyncio ? It's very new and I don't think there is a lot of people using it, but it seems to work (except the test_commands.py that break everything)

`DATABASE=postgres pytest guillotina/ --ignore=guillotina/tests/test_commands.py`

### DIFF
--- a/guillotina/commands/__init__.py
+++ b/guillotina/commands/__init__.py
@@ -17,13 +17,6 @@ import yaml
 
 
 try:
-    import uvloop  # type: ignore
-
-    uvloop.install()
-except ImportError:
-    pass
-
-try:
     import line_profiler  # type: ignore
 
     HAS_LINE_PROFILER = True
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,8 @@ CHANGELOG
 6.0.0a4 (unreleased)
 --------------------
 
-- Nothing changed yet.
+- Replace pytest-asyncio with alt-pytest-asyncio
+  [masipcat]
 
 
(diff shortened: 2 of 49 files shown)


guillotina/commands/__init__.py
```diff
@@ -16,13 +16,6 @@
 import yaml
 
 
-try:
-    import uvloop  # type: ignore
-
-    uvloop.install()
-except ImportError:
-    pass
-
```
masipcat: @bloodbare  this was causing some tests fails because it was changing the current loop.

I was about to move this to Command.run_command() but I found this: https://github.com/plone/guillotina/pull/788/files#diff-fdbd14b1d4ea412cb601947ddd0b74afL142

I'm not sure what to do

jordic:
Have you explored something like this?
https://stackoverflow.com/questions/38594203/py-test-add-marker-to-all-tests

guillotina/commands/__init__.py
vangheem: Yes, we shouldn't install this automatically I think.  It should be in the bootstrap of the loop.

I didn't think this was even necessary if you set the loop policy setting.

vangheem:
https://github.com/delfick/alt-pytest-asyncio seems nice. I'm game for moving to it.

masipcat:
I'll try this https://github.com/delfick/alt-pytest-asyncio/pull/1 as a workaround for `test_commands`. I'll test as well the @jordic solution.

PR #795, opened 2019-12-20
Improving ValidationErrors messages

masipcat:
~~This is still not complete, but I'd like your opinion on this changes, to be sure I'm doing on the right direction~~

I'd say this first iteration is ready to be tested


masipcat:
In this PR I'm introducing this changes:

* `ValidationError` have this five optional attributes: `value`, `exp_type`, `field_name`, `errors`, `constraint`.
* All exceptions derived from `ValidationErrror` should use the previous attributes (if apply).
* `ValidationError` has a method to generate the json representation (can be moved to an adapter). This way we can generate the json for exceptions with nested errors/exceptions.

What do you think @bloodbare @vangheem  ?

masipcat:
Let me change one thing before merging this

masipcat:
Now is 100% ready!

Please squash commits on merge and release a new alpha if you don't mind :P

PR #797, opened 2019-12-21
[master] Fix error with requeued async queue tasks

vangheem:


guillotina/tests/test_queue.py
```diff
@@ -87,3 +88,22 @@ def __init__(self):
 
     for job in jobs:
         assert job.func.done
+
+
+async def test_retry_on_txn_not_finished(guillotina, loop):
```
masipcat: Tests are failing becaues `loop` (from aiohttp fixtures) doesn't exist. Maybe we should rename the fixture `event_loop` to `loop` to make the migration to G6 easier

vangheem: Thanks. I keep forgetting when I cherry pick these

Issue #750, opened 2019-12-06
Empty id containers could be created


jordic:
```
POST 
/db/guillotina/
{
"@type": "Container",
"id": ""
}
```

vangheem:
Dang. That probably means you can't delete this object afterwards either

jordic:
yup! I deleted it throught pg

PR #799, opened 2019-12-22
Prevent creating containers with empty id

lferran:
Fixes https://github.com/plone/guillotina/issues/750

### DIFF
--- a/guillotina/api/container.py
+++ b/guillotina/api/container.py
@@ -21,7 +21,6 @@ from guillotina.response import ErrorResponse
 from guillotina.response import HTTPConflict
 from guillotina.response import HTTPNotFound
 from guillotina.response import HTTPNotImplemented
-from guillotina.response import HTTPPreconditionFailed
 from guillotina.response import Response
 from guillotina.utils import get_authenticated_user_id
 from typing import Optional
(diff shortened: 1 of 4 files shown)


guillotina/api/container.py
```diff
@@ -114,8 +114,12 @@ class DefaultPOST(Service):
         if "@type" not in data or data["@type"] not in app_settings["container_types"]:
             raise HTTPNotFound(content={"message": "can not create this type %s" % data["@type"]})
 
-        if "id" not in data:
-            raise HTTPPreconditionFailed(content={"message": "We need an id"})
+        try:
+            assert data["id"]
```
masipcat: I think it's better to change the assert for an if, because assert can be disabled when running Python with `-O` (https://stackoverflow.com/questions/1273211/disable-assertions-in-python)

masipcat: Also, maybe the fix should be applied here https://github.com/plone/guillotina/blob/73a4793e70531105a2560863e818b758823dc3f1/guillotina/content.py#L561 ? Is it intended make ID optional?

vangheem:
Needs flake8 updates!

guillotina/json/definitions.py
```diff
@@ -67,7 +67,11 @@
     {
         "type": "object",
         "title": "Base resource data",
-        "properties": {"id": {"type": "string"}, "@type": {"type": "string"}, "title": {"type": "string"}},
+        "properties": {
+            "id": {"type": "string", "minLength": 1},
```
lferran: here's the actual fix

lferran:
@bloodbare do you approve?

Issue #778, opened 2019-12-16
DbUsers. When Adding a new group named Managers users with the group managers become superusers


jordic:
The wired thing is tha't it needs to be set up intentional, perhaps is a best option to just create it, on addon installation.

jordic:
@lferran any thoughts? What do you think if we just create the managers group on install? This seems like a more explicit behaviour.

lferran:
yes! I'm ok with this!
We need to do the same for authenticated, and anonymous, no?

jordic:
i don't think... Authenticated and anonymous are implicit.. without session with session, on the other side, managers are just an upgrade from authenticated, no?

Issue #771, opened 2019-12-13
@addable-types and containers


jordic:
If I'm not wrong the expected behavior for `@addable-types` should be to show types, that at the same time, are installed via @addons on the container. Currently, `@addable-types` shows you all types, living on the guillotina instalation.

Example:
```
/db/guillotina 
   -> Has the package xxx addon installed that registers Type1
/db/guillotina2/@addable-types 
   return Type1... without having the addon installed
```

Is this expected?

PR #800, opened 2019-12-22
Be able to configure type as not globally addable

lferran:


guillotina/tests/test_content.py
```diff
@@ -129,6 +129,48 @@ class CustomContentType(Item):
...
+        import guillotina.tests
+
+        # Define BarType which is not globally addable
+        configure.register_configuration(
+            Folder,
+            dict(type_name="BarType", module=guillotina.tests, globally_disallowed=True),
+            "contenttype",
+        )
+
+        # Define the only content where TestType is addable
```
lferran:
```suggestion
        # Define the only content where BarType is addable
```

jordic:
So sorry, I'm not pretty sure what's on, just discovered the Interface IConstrainTYpes.. perhaps it's as easy of implementing this? 
There is also the conf setting of 'cobtainer_types'.
@vangheem?

guillotina/api/content.py
```diff
@@ -194,7 +194,9 @@ class DefaultPOST(Service):
 
         # Create object
         try:
-            obj = await create_content_in_container(self.context, type_, new_id, **options)
+            obj = await create_content_in_container(
+                self.context, type_, new_id, skip_global_constrain=False, **options
```
vangheem: How about instead of `skip_global_constrain`, we just have a way to ignore all constraints?

So just rename to `check_constraint` just like the `check_security`

guillotina/constraintypes.py
```diff
@@ -13,22 +14,31 @@ class FTIConstrainAllowedTypes:
     def __init__(self, context: IResource) -> None:
         self.context = context
 
-    def is_type_allowed(self, type_id: str) -> bool:
+    def is_type_allowed(self, type_id: str, skip_global_constrain: bool = False) -> bool:
```
vangheem: So I think with the above comment, we can remove `skip_global_constrain` right?

guillotina/content.py
```diff
@@ -101,17 +101,19 @@ def __init__(
         type_name="",
         schema=None,
         behaviors=None,
-        add_permission=DEFAULT_ADD_PERMISSION,
-        allowed_types=None,
+        add_permission: str = DEFAULT_ADD_PERMISSION,
+        allowed_types: Optional[List[str]] = None,
+        globally_disallowed: bool = False,
```
vangheem: Can we use positive naming maybe? Rename `globally_disallowed` to `globally_addable` and default to `True` then?

guillotina/content.py
```diff
@@ -567,7 +569,13 @@ def iter_schemata(obj) -> Iterator[Interface]:
 
 @profilable
 async def create_content_in_container(
-    parent: Folder, type_: str, id_: str, request: IRequest = None, check_security=True, **kw
+    parent: Folder,
+    type_: str,
+    id_: str,
+    request: IRequest = None,
+    check_security: bool = True,
+    skip_global_constrain: bool = False,
```
vangheem: rename to `check_constraint` and default to `True`?

lferran:
@vangheem done the requested changes :+1:

guillotina/api/content.py
```diff
@@ -196,7 +196,9 @@ class DefaultPOST(Service):
 
         # Create object
         try:
-            obj = await create_content_in_container(self.context, type_, new_id, **options)
+            obj = await create_content_in_container(
+                self.context, type_, new_id, check_constrain=True, **options
```
vangheem: can we rename to `check_constraints`?

guillotina/commands/testdata.py
```diff
@@ -192,6 +192,7 @@ def get_parser(self):
                 creators=("root",),
                 contributors=("root",),
                 title=page_data["title"],
+                check_constrain=False,
```
vangheem: Here too

guillotina/configure/__init__.py
```diff
@@ -174,7 +174,15 @@ def load_contenttype(_context, contenttype):
         behaviors=[resolve_dotted_name(b) for b in conf.get("behaviors", []) or ()],
         add_permission=conf.get("add_permission") or DEFAULT_ADD_PERMISSION,
         allowed_types=conf.get("allowed_types", None),
+        globally_addable=conf.get("globally_addable", True),
     )
+
+    # Populate app settings with globally addable types
+    if not factory.globally_addable:
```
vangheem: what is the purpose of this? I think we can remove right?

guillotina/content.py
```diff
@@ -569,7 +571,13 @@ def iter_schemata(obj) -> Iterator[Interface]:
 
 @profilable
 async def create_content_in_container(
-    parent: Folder, type_: str, id_: str, request: IRequest = None, check_security=True, **kw
+    parent: Folder,
+    type_: str,
+    id_: str,
+    request: IRequest = None,
+    check_security: bool = True,
+    check_constrain: bool = True,
```
vangheem: Naming here

guillotina/content.py
```diff
@@ -714,6 +723,7 @@ def get_all_behavior_interfaces(content) -> list:
         creators=creators,
         contributors=contributors,
         check_security=check_permission,
+        check_constrain=True,
```
vangheem: here too.

guillotina/configure/__init__.py
lferran: I think this was a request by @bloodbare to be compatible with `guillotina_cms` https://github.com/plone/guillotina_cms/blob/master/guillotina_cms/content/contraintypes.py#L20

guillotina/api/content.py
lferran: done

guillotina/configure/__init__.py
bloodbare: we have this functionallity on GCMS, it can be kept there. In case we move to G should be on _settings.py and no need to setdefault.

vangheem: ahh, gotcha.

vangheem: So @lferran it sounds like this can be removed then.

guillotina/content.py
```diff
@@ -569,7 +571,13 @@ def iter_schemata(obj) -> Iterator[Interface]:
 
 @profilable
 async def create_content_in_container(
-    parent: Folder, type_: str, id_: str, request: IRequest = None, check_security=True, **kw
+    parent: Folder,
+    type_: str,
+    id_: str,
+    request: IRequest = None,
+    check_security: bool = True,
+    check_constraints: bool = True,
```
lferran: just last comment before the merge, to be sure: @bloodbare did you prefer `check_constrains=False` by default on `create_content_in_container`?

vangheem: Default should be true!

lferran: cool! then we're good to merge !

jordic:
Hey, tha't super cool, but it solves the issue https://github.com/plone/guillotina/issues/778 or this https://github.com/plone/guillotina/issues/771. 

May I close them?

What should I do to restrict some types to only some folders?

lferran:
@jordic you have to define that content as `@configure.contenttype("MyType", ..., globally_addable=False)` but then allow it on those folders `@configure.contenttype("MyFolder", ..., allowed_types=["MyType"])`

jordic:
Then we could close both issues :)

PR #802, opened 2019-12-26
Fix validating None with required value

vangheem:


vangheem:
@bloodbare thoughts on this?

Basically, this fixes the scenario where you provide `None` for a value and expect it to validate the field.

Without this, for example, you can put `required=True` and it is meaningless even when you put `None` in the payload for the value. It's really not what you would expect with our implementation because if you leave it out of the payload, you'll get required error.

However, is it possible that we are considering `None` to be different than `required`?

If it is the case, the behavior seems very odd.

In any case, I need to solution here... so I'm fine implementing other suggestions .

PR #805, opened 2019-12-29
Add localroles to @available-roles

jordic:
To be able to use the permission system on gmi, we need to have an
endpoint to get declared roles for the running guillotina.
It's not enought with globalroles (these are mostly related to container
admin), we need also local roles.


vangheem:
Hi @jordic ! Looks like you have a black error. 

If you install the pre-commit hook you'll never see this problem again :)

https://github.com/plone/guillotina#getting-started-with-development

jordic:
Please, squash!

PR #807, opened 2019-12-30
fix group denormalitzation when a user is patched

jordic:
Throught the API we can patch a user and add it to a new group,
but this is not correctly denormalized on the groups side.
This fixes it adding a subscriber to the user modified event.


vangheem:
Hey @jordic there is a test failure!

jordic:
@masipcat @vangheem @bloodbare I have a wired, issue here, it's like when running the test, the db it's not cleared after each run, (or perhaps a cache not invalidated?)

guillotina/contrib/dbusers/subscribers.py
```diff
@@ -44,3 +44,18 @@
...
+    container = get_current_container()
+    user_groups = user.user_groups or []
+    for group_id in user_groups:
+        try:
+            group = await navigate_to(container, f"groups/{group_id}")
+        except KeyError:
+            raise HTTPPreconditionFailed(content={"reason": f"inexistent group: {group_id}"})
+        if user.id not in group.users:
+            group.users.append(user.id)
+            group.register()
```
masipcat: I think its missing the logic for when a group is removed in the `user_groups`. IMO would be better to implement the `IObjectBeforeModifiedEvent` and get the differences between `user.user_groups` and `payload["user_groups"]`. That way we know which groups are added and which ones are deleted.

jordic: nice idea!
Anywyay there is something that feels like so bad... 
On the current PR, why the test fails when runs in conjuntion with test_manage_groups.py, 
There is somehting contaminated on it.

jordic: o my god! I'm hitting an issue that smells like a mutable shared list.. :/

masipcat: This might be the problem: https://github.com/plone/guillotina/blob/241524086b35a016e214c0d0b8b6908ea047f586/guillotina/contrib/dbusers/content/groups.py#L26

I think it should be `defaultFactory=list` instead of `default=[]`

masipcat: Also this: https://github.com/plone/guillotina/blob/241524086b35a016e214c0d0b8b6908ea047f586/guillotina/contrib/dbusers/content/groups.py#L38

guillotina/contrib/dbusers/subscribers.py
```diff
@@ -44,3 +48,48 @@
         if group.id not in user.user_groups:
             user.user_groups.append(group.id)
             user.register()
+
+
+@configure.subscriber(for_=(IUser, BeforeObjectModifiedEvent))
```
masipcat: This should be `IBeforeObjectModifiedEvent`, right?

guillotina/contrib/dbusers/content/users.py
```diff
@@ -57,10 +57,17 @@ class IUser(IFolder, IPrincipal):
...
     username = email = name = password = None
     disabled = False
-    user_roles = ["guillotina.Member"]
-    user_groups: typing.List[str] = []
-    user_permissions: typing.List[str] = []
-    properties: typing.Dict[str, typing.Any] = {}
+    user_roles = None
+    user_groups: typing.Optional[typing.List[str]] = None
+    user_permissions: typing.Optional[typing.List[str]] = None
+    properties: typing.Optional[typing.Dict[str, typing.Any]] = None
```
masipcat: Just wondering: whats the point of defining the class attributes? These attributes are created dynamically from the content-type schema, right? If its for the mypy annotations, now that the attributes are defined in the __init__, we can define the annotations there.

(Same for Group contenttype)

jordic: Thought it's something inherited from the initial implementation, agree on clean them.

guillotina/contrib/dbusers/content/groups.py
```diff
@@ -33,9 +31,11 @@ class IGroup(IFolder):
     behaviors=["guillotina.behaviors.dublincore.IDublinCore"],
 )
 class Group(Folder):
-    name = description = None
-    user_roles: typing.List[str] = []
-    users: typing.List[str] = []
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.name = self.description = None
+        self.user_roles = []
```
lferran:
```suggestion
        self.user_roles: typing.List[str] = []
```

guillotina/contrib/dbusers/content/groups.py
```diff
@@ -33,9 +31,11 @@ class IGroup(IFolder):
...
+        self.users = []
```
lferran:
```suggestion
        self.users: typing.List[str] = []
```

guillotina/contrib/dbusers/content/users.py
```diff
@@ -55,12 +52,14 @@ class IUser(IFolder, IPrincipal):
...
-    user_groups: typing.List[str] = []
-    user_permissions: typing.List[str] = []
-    properties: typing.Dict[str, typing.Any] = {}
+    def __init__(self, *args, **kwargs):
+        self.user_groups = []
+        self.user_permissions = []
+        self.user_roles = []
+        self.properties = {}
+        self.username = self.email = self.name = self.password = None
+        self.disabled = False
```
lferran:
```suggestion
        self.disabled: bool = False
```

guillotina/contrib/dbusers/content/users.py
```diff
@@ -55,12 +52,14 @@ class IUser(IFolder, IPrincipal):
...
 )
 class User(Folder):
-    username = email = name = password = None
-    disabled = False
-    user_roles = ["guillotina.Member"]
-    user_groups: typing.List[str] = []
-    user_permissions: typing.List[str] = []
-    properties: typing.Dict[str, typing.Any] = {}
+    def __init__(self, *args, **kwargs):
+        self.user_groups = []
```
lferran:
```suggestion
        self.user_groups: typing.List[str] = []
```

lferran: and the same for the others

guillotina/contrib/dbusers/serializers.py
```diff
@@ -39,6 +40,8 @@ def __init__(self, context, request):
             self.context.email = data["email"]
         if "fullname" in data:
             self.context.name = data["fullname"]
+        if "user_groups" in data:
+            self.context.user_groups = data["user_groups"]
```
lferran: As a future improvement, maybe user groups, roles, permissions, and so on, would be nice to have them defined in the schema as a `PatchField(List)`, so we can have partial updates aswell?

guillotina/contrib/dbusers/content/users.py
jordic: What's tha point on typing this if types doesn't esxist at runtime?

guillotina/contrib/dbusers/serializers.py
jordic: Mmm.. not now..  this is a braking change that would brake guillotina react..

guillotina/contrib/dbusers/content/users.py
jordic: But also, we removed them, because it's already typed with the schema definition

jordic: On the end, interfaces are validated at runtime.. mypy types not. doesn't make sense to double type something.

guillotina/contrib/dbusers/serializers.py
vangheem: PatchField works as regular field as well.

vangheem:
Merging. I think the type hint suggestions were already done in the interface so it should be good. Thank you!

PR #809, opened 2020-01-02
Deserialization optimizations

vangheem:
## deserialize value:

Without:
```
Done capnp in 0.09863615036010742
Done protobuf in 0.9731929302215576
Done python in 3.329144239425659
Done jsonschema in 1.3023309707641602
```

with:
```
Done capnp in 0.09863615036010742
Done protobuf in 0.9224121570587158
Done python in 0.5194058418273926
Done jsonschema in 1.2855918407440186
```

So quite a bit of improvement here for python deserialization...

## deserialize content

without:

```
Test content serialization with TestContent1
Done with 1000 in 6.665909290313721 seconds
Test content serialization with TestContent6
Done with 1000 in 6.721236944198608 seconds
```

With:
```
Test content serialization with TestContent1
Done with 1000 in 2.075688362121582 seconds
Test content serialization with TestContent6
Done with 1000 in 2.0040640830993652 seconds
```


guillotina/json/deserialize_value.py
```diff
@@ -37,18 +40,32 @@ def schema_compatible(value, schema_or_field, context=None):
...
 
+def _optimized_lookup(value, field, context):
+    if field._type in (int, str, float, bool):
+        # for primitive types, all we really do is return the value back.
+        # this is costly for all the lookups
+        if not isinstance(value, field._type):
+            raise WrongType(value, field._type, field.__name__)
+        return value
+    else:
+        return schema_compatible(value, field, context)
```
masipcat: Question: do you think that it could be possible to cache the result of `get_adapter()` for given a schema or field? So we can have a dict of fields mapped to a callable (the adapter) so we don't need to call the get_adapter (that it seems what's slowing down the system)?

vangheem: In other tests that I've done, the adapter lookups are about as fast as dictionary lookups so there wouldn't be much gain there.

The real gain was actually in the adapter itself--if it was a field that didn't require any additional serialization.

I'll run more tests though. It's possible this `_optimized_lookup` isn't worth it anymore. The biggest gain in the module was removing the `field.from_unicode(value)` call which called `validate`--it was done on every sub field.

masipcat: Ok! :-)

vangheem:
@jordic @bloodbare I'm ready for review now. This has significant performance improvements; however, there are some potential breaking changes here--the size restrictions.

I can remove this breaking change and only include it in the g6 version. Or just remove all together and it's just something I do in my own code.

guillotina/json/deserialize_value.py
```diff
@@ -37,18 +40,32 @@ def schema_compatible(value, schema_or_field, context=None):
         raise ValueDeserializationError(schema_or_field, value, "Deserializer not found for field")
 
 
+def _optimized_lookup(value, field, context):
```
bloodbare: Couldn’t we just extend schema_compatible with optimized_lookup code ?

vangheem: good point!

vangheem: Actually, no we can't. We still need top-level fields to be looked up for correct implementation. For example, patch fields.

PR #813, opened 2020-01-03
Update apt install in Dockerfile

svx:
This PR improves `apt install` in the *Dockerfile*.

With the following:

- adding `no-install-recommends`
- removing `apt cache`

The result is a smaller image:

**Old:** 760MB
**NEW:** 725MB

There is a lot more we can do to make this image smaller, faster and such, but let's start small one step at the time, to make sure everything goes well :)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,9 @@ CHANGELOG
 - Add localroles to @available-roles
   [jordic]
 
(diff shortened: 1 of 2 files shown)


vangheem:
nice, thank you for taking care of this!

vangheem:
Can you also make a PR for the 5.x branch please?

svx:
Sure! Will do!!

PR #815, opened 2020-01-04
Integrate towncrier

vangheem:
See https://pypi.org/project/towncrier/

This is so we don't get merge conflicts on CHANGELOG entries.

First release will need some manual fixing.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,8 @@
 CHANGELOG
 =========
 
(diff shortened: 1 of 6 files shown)


ale-rt:
Personal experience: if you lose time rebasing/merging branches because of conflicts in the changelog it is definitely worth it.
Note: you need to be more careful when releasing (you might want to check the merged changelog)

vangheem:
I guess closing...

PR #816, opened 2020-01-07
Update Dockerfile

svx:
This PR improves apt install in the Dockerfile.

With the following:

- adding `no-install-recommends`
- removing `apt cache`

The result is a smaller image:

Old: 760MB
New: 725MB


svx:
@bloodbare master branch is already done! :) -> https://github.com/plone/guillotina/pull/813 :)

PR #818, opened 2020-01-07
[master] Deserialization optimizations

vangheem:


guillotina/tests/test_api.py
```diff
@@ -1226,11 +1226,7 @@ class TestSchema(Item):
         )
         assert status == 412
         assert resp["deserialization_errors"] == [
-            {
-                "errors": [{"field": "tags", "message": "Expected <class 'str'> but found <class 'int'>."}],
-                "field": "tags",
-                "message": "Wrong contained type",
-            }
+            {"message": "Expected <class 'str'> but found <class 'int'>.", "field": "tags"}
```
masipcat: Is there any way we can keep the nested error list?

vangheem: yes, I'll see what I can do!

masipcat: I think adding this again will do the fix https://github.com/plone/guillotina/pull/818/files#diff-b2a845fe39a48b7b8c8bbeb584780abaL85

Also in IDict deserializer

masipcat: And IObject as well

vangheem: I think my latest change fixed it if you could take a look.

masipcat: Yes, now i see the code I was referring to is here: https://github.com/plone/guillotina/pull/818/files#diff-b2a845fe39a48b7b8c8bbeb584780abaR57

masipcat: Thank you :)

PR #823, opened 2020-01-09
[5.x] Be able to start database transaction before transaction has started

vangheem:


vangheem:
We use `resolve_readcommitted` txn strategy for our storage databases.

This strategy delays starting the db transaction to just after the "before" commit hooks are run and just before it starts pushing all the data to pg.

This allows us to, in some cases, begin the actual db txn in before commit hooks or even in view code without error.

masipcat:
I'm curious.  How do you start the pg txn from a service? With storage.start_transaction() ?

vangheem:
Yes

Issue #746, opened 2019-12-02
PubSubUtility doesn't recover after a redis restart


masipcat:
The current implementation just logs the following error but doesn't reinitializes the connection.

Traceback:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/guillotina/contrib/pubsub/utility.py", line 48, in real_subscribe
    channel = await self._driver.subscribe(channel_name)
  File "/usr/local/lib/python3.7/site-packages/guillotina/contrib/redis/driver.py", line 110, in subscribe
    (channel,) = await self._pubsub_subscriptor.subscribe(channel_name)
  File "/usr/local/lib/python3.7/site-packages/aioredis/commands/pubsub.py", line 31, in subscribe
    conn.execute_pubsub(b'SUBSCRIBE', channel, *channels),
  File "/usr/local/lib/python3.7/site-packages/aioredis/connection.py", line 363, in execute_pubsub
    raise ConnectionClosedError("Connection closed or corrupted")
aioredis.errors.ConnectionClosedError: Connection closed or corrupted
```

Log:
```
Unhandled exception with pubsub. Sleeping before trying again
```

vangheem:
It is in a `while channel_name in self._subscribers:`

Shouldn't it retry?

masipcat:
Yeah. All I know right now is the guillotina pod "was failing" until we restarted the instance... I'll reproduce this locally and see what happens.

masipcat:
https://aioredis.readthedocs.io/en/latest/api_reference.html#aioredis.ConnectionsPool.execute_pubsub

> There is no auto-reconnect for Pub/Sub connection as this will hide from user messages loss.

Any idea on how to fix this on guillotina? or at least, make sure that PubSub utility doesn't goes to a unresponsive state without recovering.

masipcat:
Maybe we can try this: https://github.com/aio-libs/aioredis/issues/409#issuecomment-399094609

vangheem:
I'm not sure.

We already do this:

```
async def _listener(self, channel: aioredis.Channel):
        while await channel.wait_message():
            msg = await channel.get()
            yield msg
```

Can we try this...

- do not use the pool for pub/sub
- on any error, close the connection and create a new one

masipcat:
I think we can close this issue for now

PR #830, opened 2020-01-13
Better error handling on redis connection issues

vangheem:
@masipcat dunno if this is related to any of the issues you noticed.

We had a situation with a bunch of errors reported. I haven't been able to reproduce this issue manually though.


guillotina/contrib/redis/driver.py
```diff
@@ -23,22 +24,31 @@ def __init__(self):
...
+                        logger.error("Error initializing pubsub", exc_info=True)
+
+    @backoff.on_exception(backoff.expo, (OSError,), max_time=30, max_tries=4)
+    async def _connect(self):
+        settings = app_settings["redis"]
+        self._pool = await aioredis.create_pool(
+            (settings["host"], settings["port"]), **settings["pool"], loop=self._loop
+        )
+        self._conn = await self._pool.acquire()
+        self._pubsub_subscriptor = aioredis.Redis(self._conn)
```
masipcat: I think this changes doesn't fix what happend to us -- we restarted the redis server and guillotina wasn't able to recover. 

Looking at the code, I think that the problem is here. Since we use the same `self._conn` during the lifetime of the guillotina instance, if the connection dies doens't recover, since we don't return the connection to the pool (and can't be cleaned up). Therefore the pubsub_subscriptor.subscribre() keeps failing (as we experienced). Makes senes?

masipcat: Here is the trabeback of the error: https://github.com/plone/guillotina/issues/746#issue-531203035

vangheem: Yup, it makes sense. I just wasn't able to reproduce it by killing redis while guillotina was running and turning it back on. At least for, the connection would re-establish.

I'm going to play with it a bit more.

masipcat: I couldn't reproduce it locally as well...

vangheem: Check now @masipcat ?

PR #831, opened 2020-01-13
Fix value_deserializer() when field.key_type._type is None

masipcat:


masipcat:
> I think we should be safe and use getattr

Like this?

```
        if getattr(field, 'key_type', None) and getattr(field.key_type, '_type', None):
            if not isinstance(key, field.key_type._type):
                raise ValueDeserializationError(field, value, "Invalid key type provided")
        result[key] = _optimized_lookup(value[key], field.value_type, context)
```

PR #834, opened 2020-01-14
[master] Fix automatic type conversion on nested fields

vangheem:


vangheem:
@masipcat so I don't know about this anymore. Did you see the problem I noticed? (this is a problem from even before I made the optimizations)

PR #836, opened 2020-01-15
Run _update_from_pytest_markers() after configuring db settings

masipcat:
The motivation behind this change is running a test with autovacuum off


masipcat:
Now its ready

PR #840, opened 2020-01-16
Fix query params validation for arrays

lferran:


guillotina/api/service.py
```diff
@@ -73,15 +86,14 @@ def _validate_parameters(self):
...
-                        try:
-                            value = int(value)
-                        except ValueError:
-                            try:
-                                value = float(value)
-                            except ValueError:
-                                pass
+                    if parameter["schema"].get("type") == "array":
+                        value = data.getall(name)
+                        if parameter["schema"]["items"].get("type") == "number":
```
vangheem: I believe `items` in json schema can be empty and it's supposed to default to `string` type
```suggestion
                        if parameter["schema"].get("items", {}).get("type") == "number":
```

PR #842, opened 2020-01-16
[5.x] Fix query params validation for arrays

lferran:


vangheem:
Hey @lferran! There is a test failure that I think is fixed by https://github.com/plone/guillotina/pull/841

Could you review maybe?

lferran:
ouch, it's because my endpoint does not have a permission defined:

```
     desc += f"permission: {service_def['permission']}"
KeyError: 'permission'
```

PR #843, opened 2020-01-16
[master] Better error handling on redis connection issues

vangheem:


guillotina/contrib/redis/driver.py
```diff
@@ -107,7 +118,18 @@ def pool(self):
     async def subscribe(self, channel_name: str):
         if self._pubsub_subscriptor is None:
             raise NoRedisConfigured()
-        (channel,) = await self._pubsub_subscriptor.subscribe(channel_name)
+        try:
+            (channel,) = await self._pubsub_subscriptor.subscribe(channel_name)
+        except aioredis.errors.ConnectionClosedError:  # pragma: no cover
+            # closed in middle
```
masipcat: I'd add a log message that we are going to close the pool, because after closing the pool other subscribe will crash (I guess...), so we know why all the subscribers crash

PR #846, opened 2020-01-17
Fix command 'create'

masipcat:


vangheem:
How come this is necessary? Create is for cookie cutter. We don't need app in that case right?

masipcat:
It's needed because I moved the app.startup() to the AsgiApp to make the initialization work when running guillotina using a asgi server directly (without calling the command 'server').

masipcat:
But I think we could rethink this part. Also how we manage the settings/config_file and the app lifecycle.

vangheem:
gotcha

PR #847, opened 2020-01-19
Add alpha version of react-gmi

jordic:
Not sure if this should be on the default guillotina distribution,
or perhaps it's a best approach to just distribute it as
pypi package.

something like:
`pip install guillotinaweb_reactgmi`

Anyway let's discuss it.


bloodbare:
I like the aproach of embedded fronts on Guillotina out of the box

bloodbare:
Just as a comment: I don't a problem on having executioner and guillotina_react out of the box. Executioner is a good NG example and GReact is a good application.

PR #849, opened 2020-01-20
Add attribute 'view_dotted_name' in request object

masipcat:
@vangheem @bloodbare I open this PR to discuss how to access some request attributes from a middleware. guillotina_prometheus needs `request.found_view` to label the observed metric, but some request attributes are cleared after the traversal. I see two solutions:

1. Add `request.view_dotted_name` for the use-case of prometheus
2. Move the request cleanup logic after this line: https://github.com/plone/guillotina/blob/master/guillotina/asgi.py#L113

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,9 @@ CHANGELOG
 6.0.0a7 (unreleased)
 --------------------
 
+- Add attribute 'view_dotted_name' in request object
+  [masipcat]
+
 - Fix validating array params in query parameters [lferran]
 
 - Add open api tests and fix ones that do not pass tests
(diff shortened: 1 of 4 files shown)


masipcat:
The problem is that the `found_view` is only available during the traversal/service execution because it's cleared here: https://github.com/plone/guillotina/blob/cf975d9cba569c90e7f4a3a54adcf6562d49dbca/guillotina/traversal.py#L304
https://github.com/plone/guillotina/blob/cf975d9cba569c90e7f4a3a54adcf6562d49dbca/guillotina/traversal.py#L232-L243
Is it ok to move the `_clean_request` after running the middlewares? https://github.com/plone/guillotina/blob/master/guillotina/asgi.py#L113

masipcat:
And `request.view_dotted_name` wouldn't be necessary anymore

vangheem:
Yup, definitely. We can move the clean request until the end of all middlewares!

masipcat:
I open a new PR: https://github.com/plone/guillotina/pull/849

PR #851, opened 2020-01-20
Remove unused methods in Response

masipcat:


guillotina/response.py
```diff
@@ -124,21 +122,6 @@ def __repr__(self) -> str:
...
-    def __getitem__(self, key: str) -> Any:
-        return self._state[key]
-
-    def __setitem__(self, key: str, value: Any) -> None:
-        self._state[key] = value
-
-    def __delitem__(self, key: str) -> None:
-        del self._state[key]
-
-    def __len__(self) -> int:
```
masipcat: since `_state` is always empty, `resp.__len__()` returns always 0, and makes `bool(resp)` return `False`. This is a problem when doing `if resp: ...`

PR #852, opened 2020-01-20
Improvements in contrib.dbusers

masipcat:
Things I'm working on:

1. Ensure that there are subscribers of events `Added`, `Modified` and `Removed` for `User` and `Group` (this is a continuation of @jordic PR)
2. Implement the `IGroups` utility for dbusers. The problem I see so far is that `get_principal()` needs to be a coroutine so we can load the group from groups folder and fill the `GuillotinaGroup` with the roles and permissions. (see https://github.com/plone/guillotina/pull/852/files#r368751920)


guillotina/tests/dbusers/test_security.py
```diff
@@ -0,0 +1,79 @@
...
+        resp, status_code = await requester(
+            "POST", "/db/guillotina/@login", data=json.dumps({"username": "007", "password": "secret"})
+        )
+        assert status_code == 200
+
+        # User should have access to its own folder
+        _, status_code = await requester(
+            "GET", "/db/guillotina/secrets/@top-secret", token=resp["token"], auth_type="Bearer"
+        )
+        assert status_code == 200
```
masipcat: This is failing with status 401

guillotina/contrib/dbusers/groups.py
```diff
@@ -0,0 +1,20 @@
...
+from guillotina.auth.groups import GuillotinaGroup
+from guillotina.interfaces import IGroups
+
+
+@configure.utility(provides=IGroups)
+class GroupsUtility:
+    """ Class used to get groups. """
+
+    def get_principal(self, ident, principal=None):
+        # import pdb; pdb.set_trace()
```
lferran: you probably need some db lookups here right?

something like
```python
try:
    group = await navigate_to(get_current_container(), f"groups/{ident}")
except KeyError:
    return None ?
return GuillotinaGroup(
    ident,
    roles=group.roles,
    permissions=group.permissions
)
```

guillotina/contrib/dbusers/groups.py
```diff
@@ -0,0 +1,20 @@
+from guillotina import configure
+from guillotina.auth.groups import GuillotinaGroup
+from guillotina.interfaces import IGroups
+
+
+@configure.utility(provides=IGroups)
+class GroupsUtility:
```
lferran: IMHO, from just looking at the implementation, it is a bit hard to understand what the utility is intended for (I mean the one that is already defined in guillotina.auth.groups.py)

What about renaming it to `IGetGroupsUtility` or `GroupsSearch`?

Also maybe typing `principal: Optional[IPrincipal]` could help?

guillotina/contrib/dbusers/groups.py
masipcat: Exactly this, but I need to make get_principals async, and I'm not sure this is a change we can make

guillotina/contrib/dbusers/groups.py
masipcat: I'd say that the interface should be similar to the `auth_user_identifier` so we can get a group doing `async find_group(id)`

masipcat:
One more thing: the purpose of `@user_info` (https://github.com/plone/guillotina/blob/master/guillotina/contrib/dbusers/services/users.py#L29) is the same as `@user` (https://github.com/plone/guillotina/blob/82d76af40332c2017fc7ca6278a542866d2b63a4/guillotina/api/user.py)?

masipcat:
> IMHO we need a group cache that populates async groups and keeps it.

Something like `load_groups()` in the `IGroups` utility? and then we load the groups for each request (to keep the roles/permissions of the groups always up to date, otherwise I'm not sure how we can keep the "cached" groups synced in all instances). Where would be the best place to call `load_groups()`?

Also, what's the purpose of this code in `IGroups.get_principal()`:
```python
 if principal is not None:
            try:
                cache = principal._groups_cache
            except AttributeError:
                cache = principal._groups_cache = {}
            if ident not in cache:
                cache[ident] = GuillotinaGroup(ident)
            return cache[ident]
```
If we add the cache dictionary as a IGroups attribute, then we don't need the `principal._groups_cache`, right?

bloodbare:
To the first question, I would say that get_user logic (that retrieved and returns the GuillotinaUser) should check if user groups are loaded on memory cache and if not load them.

For the second question, yes, we don't need

masipcat:
I'll do a first implementation so we can discuss the details ;)

guillotina/tests/dbusers/test_security.py
masipcat: Solved

guillotina/contrib/dbusers/groups.py
masipcat: I added the type annotation

guillotina/auth/utils.py
```diff
@@ -44,11 +44,16 @@ def set_authenticated_user(user):
...
+        user = root.root_user
+    else:
+        for identifier in app_settings["auth_user_identifiers"]:
+            user = await identifier().get_user(token)
+            if user is not None:
+                break
+    if user:
+        groups = get_utility(IGroups)
+        await groups.load_groups(user.groups)
+    return user
```
masipcat: @bloodbare As you said, we load the groups needed by the user

masipcat: @bloodbare and regarding your comment 
> if user groups are loaded on memory cache and if not load them.

This would cause having a stale permission/roles in the cache when the group is modified.
IMHO the cache can be reimplemented in a custom IGroups in the case of advanced users that need more performance.

masipcat:
> Wouldn't it be better to add an interface to GuillotinaUser so IGroups could be an adapter of IGuillotinaUser?

Can you provide an example? I think this doesn't solve the need of loading groups asynchronously.

> More cents, could if be done on the identifier instead of the auth general system ?

I guess we can move the `IGroups.load_groups()` to `auth_user_identifiers.get_group()` but we still need a way keep the obj groups that the user needs in memory so IGroups.get_principal() can access them.

bloodbare:
the key is to use https://github.com/plone/guillotina/blob/master/guillotina/auth/utils.py#L49

its an async function that depends on the identifiers loaded. On the Identifier we load the user and its groups on memory. Then at security policy IGroups does an adapter lookup of the user object to an adapter that populates groups inside the user object

jordic:
> I'll review in deep later, just two cents by now. Wouldn't it be better to add an interface to GuillotinaUser so IGroups could be an adapter of IGuillotinaUser?

I don't fully got why... IGroups, seems to me more or less a way of discovering which groups are available on the system.. (It should be something related to the container, no?) As a generic utility, everyone could roll it's own, For dbusers, seems like the easy path.

(Why we want to adapt IGuillotinaUser?) Which relation they have?

On the other side, I want to understand why mixing async and security policy could be dangerous? (Are you thinking on coroutines that never gots executed, or something like... )... Ensuring principals on a system seems like an async operation (it relays on network, db, ldap, or whatever... )

masipcat:
> the key is to use https://github.com/plone/guillotina/blob/master/guillotina/auth/utils.py#L49 its an async function that depends on the identifiers loaded. On the Identifier we load the user and its groups on memory.

Ok, this would work, the only thing I'm not sure is where is the best place to have the "groups in memory". In a global dict? As an attribute of the user, something similar to https://github.com/plone/guillotina/blob/82d76af40332c2017fc7ca6278a542866d2b63a4/guillotina/auth/groups.py#L26-L29 ?

>  Then at security policy IGroups does an adapter lookup of the user object to an adapter that populates groups inside the user object

And then the adapter could return the value of `user._cached_groups[ident]`, right?

Is this what you mean?

bloodbare:
> I don't fully got why... IGroups, seems to me more or less a way of discovering which groups are available on the system.. (It should be something related to the container, no?) As a generic utility, everyone could roll it's own, For dbusers, seems like the easy path.
> 
> (Why we want to adapt IGuillotinaUser?) Which relation they have?

Each identifier provided on config.yaml means a provider of identification of principals. Each one of them may get users from DB/local/API/OAuth.... Each one of this identifiers may define principals for the groups they belong to, for example, dbusers users and groups belong to a DB. Of course there is the use case were principals for groups are on one source, and users are on multiple sources. I agree that we would not cover that use case in case we adapt user and not a IGroups utility. If then we want to cover that use case we need to provide a groups_provided_identifier that is called to get the groups. I'm ok on not covering that use case and assuming that groups come from the same provider that provides the user.

> 
> On the other side, I want to understand why mixing async and security policy could be dangerous? (Are you thinking on coroutines that never gots executed, or something like... )... Ensuring principals on a system seems like an async operation (it relays on network, db, ldap, or whatever... )

Security policy is one of the most complex/repeated/cpu_intense codes on G. Adding async on it may lead to a DB query which breaks the principal that all needs to be already on memory when you check security.

bloodbare:
> Is this what you mean?

Yes, I think its clean and does not break actual flow. About performance for  groups some ideas:

- Using catalog instead of db queries
- Storing on user
- Cache by container

masipcat:
> Yes, I think its clean and does not break actual flow.

@bloodbare with what you said, the IGroups utility would be:
```python
@utility(IGroups)
class GroupsUtility:
    def get_principal(self, ident, principal=None):
         return IGroupInUser(principal)(ident)
```
This would fail when `principal` is none (and there are some cases). I think we need to find another way

guillotina/contrib/dbusers/users.py
```diff
@@ -28,4 +30,16 @@ class DBUserIdentifier:
...
+            user._groups_cache = {}
+
+        # load groups
+        for ident in user.groups:
+            try:
+                group = await navigate_to(container, f"groups/{ident}")
+            except KeyError:
+                continue
+
+            user._groups_cache[ident] = GuillotinaGroup(ident, roles=group.roles)
```
jordic: This is so wired to be part of a public API, no? At least, what we should answer to the question how you can provide your own auth API?

masipcat: Yes, I agree. This is something I want to discuss why @bloodbare . The point of this change is not breaking compatibility with existing custom implementations, but I think this should go in a dedicated method like `async get_groups(user)` in the user_identifier.

jordic: We can't loose the main focus of the issue, we are touching this part, because we want to clarify the "extensible" API to roll our own auth/identification system. and on the way we found the this API it's not complete (ready) on the dbusers referential implementation.

masipcat: I updated `IPrincipal` to make explicit the field `_groups_cache`. Also, we don't break the existing implementation of IGroups that was using the `_groups_cache`.

guillotina/auth/groups.py
```diff
@@ -21,12 +24,9 @@ def __init__(self, ident):
...
     """ Class used to get groups. """
 
-    def get_principal(self, ident, principal=None):
+    def get_principal(self, ident: str, principal: typing.Optional[IPrincipal]) -> IPrincipal:
         if principal is not None:
-            try:
-                cache = principal._groups_cache
-            except AttributeError:
-                cache = principal._groups_cache = {}
+            cache = principal._groups_cache
```
bloodbare: Can we keep the try/except?

masipcat: done

PR #853, opened 2020-01-22
[5.x] Correctly bubble http errors for file downloads

vangheem:


guillotina/files/manager.py
```diff
@@ -231,26 +232,30 @@ def __init__(self, context, request, field):
                     )
                 await download_resp.write(chunk)
                 await download_resp.drain()
+        except Response:  # pragma: no cover
+            http_exception = True
+            raise
```
masipcat: Why we don't want to raise all Exceptions?

vangheem: We do, this is to prevent the finally block from starting a response with the incorrect content length!

masipcat: Hmm, we only do for exceptions of type `Response` not all `Exceptions`

guillotina/files/manager.py
```diff
@@ -231,26 +232,31 @@ def __init__(self, context, request, field):
...
-                    filename,
-                    content_type,
-                    size,
-                    extra_headers,
-                    range_supported=range_supported,
-                    **kwargs,
-                )
-            await download_resp.write_eof()
-            return download_resp
+            if not http_exception:
```
masipcat: Also, this could be implemented using try...except...else so we only execute the following code if no exception ocurred

guillotina/files/manager.py
vangheem: gotcha, I think you're right.

guillotina/files/manager.py
vangheem: yes, let me refactor.

PR #860, opened 2020-02-01
[master] Be able to customize pg db in test fixtures

vangheem:


vangheem:
same as here: https://github.com/plone/guillotina/pull/859/files

PR #861, opened 2020-02-03
Registration and reset flow for users

bloodbare:


guillotina/contrib/dbusers/content/users.py
```diff
@@ -85,6 +88,15 @@ def groups(self):
     def _properties(self):
         return {}
 
+    async def set_password(self, new_password, old_password=None):
+        if old_password is not None:
+            valid = check_password(self.password, old_password)
```
jordic: Do we need to add something to define some password constraints, like:
minium length, Caps chars, lower cars, numbers... (The basic one could be just a minium length).

bloodbare: I don't think that is the scope of this PR. Right now can be guided on the frontend and not enforced on the backend.

bloodbare: I've added a validation schema for validation process to reset password can be validated and minlength can be set

guillotina/api/login.py
```diff
@@ -84,3 +87,253 @@ class Refresh(Service):
...
+            try:
+                await actual_user.set_password(
+                    data.get('new_password', None),
+                    old_password=data.get('old_password', None)
+                )
+            except AttributeError:
+                raise HTTPNotAcceptable()
+        else:
+            # We validate with recaptcha
+            validator = RecaptchaValidator()
```
masipcat: I think this should be a utility that can be customized and maybe the default implementation could be a `return True`

guillotina/api/login.py
```diff
@@ -84,3 +87,253 @@ class Refresh(Service):
...
+    summary="Register Users",
+    allow_access=True,
+)
+class RegisterUsers(Service):
+    async def __call__(self):
+        allowed = app_settings.get('allow_register', False)
+        if allowed is False:
+            raise HTTPUnauthorized(content={"text": "Not allowed registration"})
+
+        validator = RecaptchaValidator()
```
masipcat: Same here

guillotina/contrib/email_validation/process.py
```diff
@@ -0,0 +1,111 @@
...
+from guillotina.component import get_utility
+from guillotina.response import HTTPPreconditionFailed
+from guillotina.utils import get_registry
+from guillotina import app_settings
+from guillotina.event import notify
+from guillotina.events import ValidationEvent
+import jsonschema
+
+
+async def start(
```
masipcat: Can we move this to an utility with a defined Interface?

guillotina/api/login.py
```diff
@@ -84,3 +87,253 @@ class Refresh(Service):
...
+                }
+            }
+        }
+    },
+    summary="Info Access",
+    allow_access=True,
+)
+class InfoAccess(Service):
+    async def __call__(self):
+        validator = RecaptchaValidator()
```
masipcat: Same

guillotina/api/login.py
bloodbare: I feel that this should be a parameter on the service view to protect the view with a CSRF like aproach. This is a quick solution that should be evolvoved to a more generic way. I think its out of the scope of this PR to provide more options that Recaptcha.

masipcat: At least would be great to be customizable. I'm ok with Recaptcha as a default implementation as long as I can replace it if I want :P

bloodbare: Right now you can disable it if you don't have the recaptcha param on settings. Looking forward for a PR to define Recaptcha/CSRF/generalized validation on view settings!

masipcat: Oh, I see, if it's not defined returns true

Issue #801, opened 2019-12-26
IMultiAttachment. Delete attachment


jordic:
Seems like the multiattachment behavior doesn't allow to delete files.
It could be awesome to just have the endpoint

`DELETE context/@upload/field/file`

On the gmi, I have a widget like this for IMultiAttachment:

![Captura de pantalla de 2019-12-26 12-03-14](https://user-images.githubusercontent.com/64216/71473573-cb457480-27d7-11ea-9f56-319f3030bdf9.png)

The feature, will allow to remove keys from the behavior

vangheem:
Definitely! I agree.

masipcat:
I see that `IFileStorageManager` doesn't defines a `delete()` method. Can we add it to the interface?

I can take care of this issue if you both agree.

vangheem:
yes please, we need this.

PR #863, opened 2020-02-03
Implemented endpoint @delete for IAttachments and IMultiAttachments

masipcat:
Closes https://github.com/plone/guillotina/issues/801

### DIFF
--- a/guillotina/api/files.py
+++ b/guillotina/api/files.py
@@ -4,6 +4,7 @@ from guillotina.api.content import DefaultOPTIONS
 from guillotina.api.service import DownloadService
 from guillotina.api.service import TraversableFieldService
 from guillotina.component import get_multi_adapter
+from guillotina.exceptions import FileNotFoundException
 from guillotina.interfaces import IAsyncBehavior
 from guillotina.interfaces import IFileManager
 from guillotina.interfaces import IResource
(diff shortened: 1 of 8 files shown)


guillotina/api/files.py
```diff
@@ -110,6 +111,34 @@ class UploadFile(TraversableFieldService):
...
+    context=IResource,
+    method="POST",
+    permission="guillotina.DeleteContent",
+    name="@delete/{field_name}",
+    parameters=[{"in": "path", "name": "field_name", "required": True, "schema": {"type": "string"}}],
+    **_traversed_file_doc("Update the content of a file"),
+)
+@configure.service(
+    context=IResource,
+    method="POST",
```
jordic: post, why?
it should be DELETE

masipcat: I'm ok with changing to DELETE. I wasn't in this case because we are not deleting the entire resource, we are doing a PATCH (I guess is the same reason `@upload` is a PATCH instead of POST)

jordic: Mmmm... The file is a resource child of...

masipcat: Well the file is stored in a field of the resource, is part of the resource. Anyway, I changed it to DELETE

masipcat:
I like it. We can provide this new api alongside with the current one.

PR #864, opened 2020-02-04
Validate POST @sharing too

lferran:


lferran:
put is already `validate=True`

vangheem:
5.3.21 released

PR #867, opened 2020-02-05
[5x.] Fix asyncpg integration with connection leaks on timeout

vangheem:


requirements.txt
```diff
@@ -2,7 +2,7 @@ Cython==0.29.6
 aiohttp>=3.0.0,<3.6.0;python_version<"3.8"
 aiohttp>=3.6.0,<4.0.0;python_version>="3.8"
 async-timeout==3.0.0
```
masipcat: I think async-timeout is not needed anymore

PR #869, opened 2020-02-06
Asyncpg pool implementation tweaks

vangheem:
1) expose more pool configuration
2) close should be asyncio task and should never timeout

### DIFF
--- a/guillotina/db/storages/pg.py
+++ b/guillotina/db/storages/pg.py
@@ -801,13 +801,13 @@ WHERE tablename = '{}' AND indexname = '{}_parent_id_id_key';
     async def _close(self, con):
         try:
             await self.pool.release(con, self._conn_release_timeout)
-        except asyncpg.exceptions.InterfaceError as ex:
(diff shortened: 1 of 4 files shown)


guillotina/db/storages/pg.py
```diff
@@ -796,11 +796,29 @@ def objects_table_name(self):
     async def open(self):
         return await self.pool.acquire(timeout=self._conn_acquire_timeout)
 
-    async def close(self, con):
+    async def _close(self, con):
         try:
-            await shield(self.pool.release(con, timeout=1))
-        except (asyncio.CancelledError, asyncio.TimeoutError, asyncpg.exceptions.ConnectionDoesNotExistError):
-            log.warning("Exception on connection close", exc_info=True)
+            await self.pool.release(con)
```
masipcat: Why not keeping the timeout with a bigger number (like 30s, 60s, ...) ? to prevent this method running indefinitely (therefore, the aiotask) is something goes bad

vangheem: sure, okay

Issue #866, opened 2020-02-04
G6 not converting Response to ASGIResponse


masipcat:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/uvicorn/protocols/http/h11_impl.py", line 385, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/usr/local/lib/python3.7/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File "/usr/local/lib/python3.7/site-packages/uvicorn/middleware/asgi2.py", line 7, in __call__
    await instance(receive, send)
  File "/usr/local/lib/python3.7/site-packages/guillotina/asgi.py", line 42, in run_asgi2
    return await self.real_asgi_app(scope, receive, send)
  File "/usr/local/lib/python3.7/site-packages/guillotina/asgi.py", line 50, in real_asgi_app
    return await self.handler(scope, receive, send)
  File "/usr/local/lib/python3.7/site-packages/guillotina/asgi.py", line 113, in handler
    if not resp.prepared:
AttributeError: 'HTTPUnauthorized' object has no attribute 'prepared'
```

PR #870, opened 2020-02-06
Moved the ASGI logic from ASGIResponse and ASGISimpleResponse to class Response

masipcat:
Closes https://github.com/plone/guillotina/issues/866

### DIFF
--- a/guillotina/asgi.py
+++ b/guillotina/asgi.py
@@ -112,6 +112,7 @@ class AsgiApp:
 
         if not resp.prepared:
             await resp.prepare(request)
+            await resp.send_body()
 
         self._cleanup(request, resp)
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,8 @@ CHANGELOG
 6.0.0a10 (unreleased)
 ---------------------
 
-- Nothing changed yet.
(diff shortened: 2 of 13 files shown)


guillotina/asgi.py
```diff
@@ -112,6 +112,7 @@ def build_middleware_stack(self, settings):
 
         if not resp.prepared:
             await resp.prepare(request)
+            await resp.send_body()
```
bloodbare: Shouldn't it check if body has started on a different if, may it be prepared and not sent?

masipcat: My idea was: if you call `prepare()` in the service (like in the download endpoints), you know what you are doing and you are responsible of managing the response flow. The `send_body()` sends the content response.body to the asgi server, but the response.body could be empty in case of streaming responses.

masipcat: I think the best solution/fix is adding:

```python
if not resp.eof:
    await resp.send_eof()
```

(this ensures the asgi server doesn't wait for the response data indefinitely)

@bloodbare agree?

bloodbare: agree, websockets will not reach this point until they are done, isn't it ?

masipcat: Exactly

PR #875, opened 2020-02-07
docs(sphinx): update copyright date

svx:
This PR updates the "copyright date" in the footer to 2020.

This is not adding value to the docs as in content and it is maybe less important for developers but it adds a general value :)

Sorry I was not able to resist 😄 .

Screenshot 2020-02-07 at 15 12 36

The date in the image shown above  ⬆️ will change to 2020.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -61,7 +61,7 @@ master_doc = "index"
 
 # General information about the project.
 project = "Guillotina"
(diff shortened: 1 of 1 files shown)


vangheem:
That could be generated. It's in python!

svx:
@vangheem Yes, I had the same thought, if I remember right, we did that in the Plone docs, years back with the old setup 😄

PR #891, opened 2020-02-12
 Improve performance of bucket dict field

vangheem:
Using bisect instead of index on large-ish lists gives us almost 10x performance.

`guillotina/tests/test_annotations.py done in 17.738866090774536`

vs

`guillotina/tests/test_annotations.py done in 1.9164221286773682`

### DIFF
--- a/guillotina/fields/annotation.py
+++ b/guillotina/fields/annotation.py
@@ -277,12 +277,11 @@ class BucketDictValue:
             # get annotation for this key again as it might be the new annotation
             annotation = await self.get_annotation(context, key)
 
-        if key in annotation["keys"]:
+        insert_idx = bisect.bisect_left(annotation["keys"], key)
+        if len(annotation["keys"]) < insert_idx and annotation["keys"][insert_idx] == key:
             # change existing value
-            idx = annotation["keys"].index(key)
-            annotation["values"][idx] = value
+            annotation["values"][insert_idx] = value
         else:
-            insert_idx = bisect.bisect_left(annotation["keys"], key)
             annotation["keys"].insert(insert_idx, key)
             annotation["values"].insert(insert_idx, value)
             _, bucket = self._find_bucket(key)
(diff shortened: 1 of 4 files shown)


guillotina/tests/test_annotations.py
```diff
@@ -94,3 +96,17 @@
         assert [k async for k in bucket.iter_keys(ob)] == [str(i) for i in _range]
         assert [v async for v in bucket.iter_values(ob)] == [i for i in _range]
         assert [(k, v) async for k, v in bucket.iter_items(ob)] == [(str(i), i) for i in _range]
+
+
+async def _test_bucket_dict_value_many_values(dummy_guillotina):
```
masipcat: test is commented?

guillotina/fields/annotation.py
```diff
@@ -277,25 +277,33 @@ def get_annotation_name(self, bucket_id: str) -> str:
             # get annotation for this key again as it might be the new annotation
             annotation = await self.get_annotation(context, key)
 
-        if key in annotation["keys"]:
+        insert_idx = bisect.bisect_left(annotation["keys"], key)
+        if len(annotation["keys"]) < insert_idx and annotation["keys"][insert_idx] == key:
```
Qiwn: shouln't be `len(annotation["keys"]) > insert_idx`?

vangheem: You're right!

guillotina/tests/test_annotations.py
vangheem: It's just to test speed. All functionality is already tested.

guillotina/fields/annotation.py
```diff
@@ -277,25 +277,33 @@ def get_annotation_name(self, bucket_id: str) -> str:
...
         annotation.register()
 
+    def __index(self, annotation, key):
+        idx = bisect.bisect_left(annotation["keys"], key)
+        try:
+            if annotation["keys"][idx] == key:
+                return idx
+        except IndexError:
+            pass
+        raise ValueError(key)
```
bloodbare: Shouldn't we provide BBB ?

vangheem: What would be bbb be needed for?

vangheem: we are only changing lookups from [].index to bisect

bloodbare:
```
>>> annotations = ['myannotation1', 'myannotation5', 'myannotation3']
>>> bisect.bisect_left(annotations, 'myannotation3')
1
>>> annotations.index('myannotation3')
2
```

masipcat: image

vangheem: @bloodbare I don't understand. This is internal implementation for inserting/updating. No apis have changed.

PR #895, opened 2020-02-13
Add `clear` patch field operation and provide `max_ops` property to patch fields

vangheem:


vangheem:
@bloodbare are you okay with this implementation?

PR #897, opened 2020-02-13
[master] Add `clear` patch field operation and provide `max_ops` property to patch fields

vangheem:


vangheem:
It should be increased. Everything is covered and I added extra tests for things not covered. I can check later but maybe the report is wrong.

bloodbare:
> It should be increased. Everything is covered and I added extra tests for things not covered. I can check later but maybe the report is wrong.

Thanks why I was commenting, I saw a red coverage but I can't see in code why

vangheem:
I don't understand the report. If you look at https://codecov.io/gh/plone/guillotina/pull/897/diff -- with what I touched, I should have better coverage...

vangheem:
I wonder if it does a bad job handling cherry pick? I'm merging. The reports everywhere on codecov do not make sense.

PR #900, opened 2020-02-17
docs(training): update note to warning

svx:
This PR updates the info box from a **Note** to a **Warning**.

The intention here is that this is important to know, and because that we want to highlight that in a different color than we do with a note :)

### DIFF
--- a/docs/source/training/configuration.md
+++ b/docs/source/training/configuration.md
@@ -44,7 +44,7 @@ docker run \
 
 
 ```eval_rst
-.. note::
+.. warning::
    This particular docker run command produces a volatile database.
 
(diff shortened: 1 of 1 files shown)


tisto:
@svx just out of curiosity. Are you (or Guillotina) using conventional commit messages? Just asking because Volto is going that direction...

svx:
@tisto I do. I started with that a while back.
Me personally, it helps a lot, because it is now easier for me to find things in my git log/history, and because in some projects I am auto generating changelogs and release summaries from git commits.

Besides all that it calms me down and gives my mind some peace, because I have now structured, unified git commits/logs.

I even heard that some folks using it in combination with webhooks to run certain CI checks only according to commit messages, maybe I will try that next :)

PR #906, opened 2020-02-18
Fixing some bugs

bloodbare:


guillotina/api/ws.py
```diff
@@ -189,4 +189,4 @@ class WebsocketsView(Service):
                     await tm.abort(txn=txn)
 
         logger.debug("websocket connection closed")
-        return {}
+        await ws.close()
```
masipcat: the ws.close() is already called in L179. Mabye L179 should be a break

PR #922, opened 2020-02-21
Handle error when "None" value provided for behavior data

vangheem:


guillotina/json/deserialize_content.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, context, request):
         # do behavior first in case they modify context values
         for behavior_schema, behavior in await get_all_behaviors(self.context, load=False):
             dotted_name = behavior_schema.__identifier__
-            if dotted_name not in data:
+            if not data.get(dotted_name):
```
masipcat: What should happen on `IDublinCore.__identifier__: {},` ?

vangheem: The reason this would error is because it would try and pull field data out of a `None` value instead.

If the value is `{}` there isn't any field data to deserialize anyways so it should be fine.

Issue #920, opened 2020-02-21
Docker Image Failing experience


bloodbare:
- Would be great that a default configuration is not needed and there is a default command on the docker image to start a dummy configuration.

- Once it starts with a dummy configuration accessing any jsapps fails. I think is related with the headers sent by ASGI server.

To reproduce:

Running:

`docker run -p 8080:8080 -it plone/guillotina g -c /app/config-dummy.yaml`

Accessing:

`http://localhost:8080/+manage/`

masipcat:
> Once it starts with a dummy configuration accessing any jsapps fails. I think is related with the headers sent by ASGI server.


Yes, we are sending application/octet-stream instead of text/html. I'll fix it later

PR #923, opened 2020-02-22
Reverted "Replaced Response.content_{type,length} with Response.set_content_{type,length}". Using setter to avoid breaking `Response.content_{type,length} = ...`

masipcat:
Closes https://github.com/plone/guillotina/issues/920


masipcat:
This PR fixes this:

https://github.com/plone/guillotina/blob/67b221f811732b6eaaacd981bc0fbbcd65e79cd8/guillotina/api/files.py#L59

PR #930, opened 2020-02-27
Changes in ICatalogUtility, DefaultSearchUtility and @search endpoints

masipcat:


masipcat:
I'm fixing guillotina_elasticsearch tests and I prefer to not merge this until I'm sure this changes works and we don't need to change it again ;)

PR #933, opened 2020-03-02
React gmi update v 0.4.0

jordic:


jordic:
@bloodbare If the size is a problem, we can discard the source map files.

PR #916, opened 2020-02-20
github action - publish docker images

psanlorenzo:
#911

Edit:
Setting secrets for dockerhub authentication. (Repository settings -> secrets)
image
More info: [Github docs](https://help.github.com/es/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets)

PR #935, opened 2020-03-03
Use github actions instead of travis

lferran:
Fixes https://github.com/plone/guillotina/issues/907


.travis.yml
```diff
@@ -1,79 +0,0 @@
...
-  script: bash docker_push
-  on:
-    tags: true
-
-# safelist
-branches:
-  only:
-  - master
-  - 4.x
-  - 5.x
```
lferran: Do you guys know why was this needed?
Do we need/want to port it to github actions aswell?

vangheem: I think it was for the docker auto deploy. It will only do it for those branches.

lferran: ok! is it something we still want? do I need to also implement this with github actions? If so, I would need more info.

vangheem: Yes, we need to auto deploy still to docker hub!

lferran: I'm confused. Maybe what @bloodbare says he configured already does the job?
https://github.com/plone/guillotina/pull/916#issuecomment-589156220

bloodbare: Yes we can remove it

lferran:
@bloodbare @vangheem @masipcat finally made it! ready to review and merge.
orders of magnitude quicker than travis :tada:

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,83 @@
...
+        if: env.DATABASE == 'postgres' || env.DB_SCHEMA == 'public'
+        uses: actions/checkout@v2
+
+      - name: Setup Python
+        if: env.DATABASE == 'postgres' || env.DB_SCHEMA == 'public'
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install the package
```
bloodbare: Why we need to reinstall ?

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,83 @@
...
+        database: ["DUMMY", "postgres", "cockroachdb"]
+        db_schema: ["custom", "public"]
+
+    # Set environment variables
+    env:
+      DATABASE: ${{ matrix.database }}
+      DB_SCHEMA: ${{ matrix.db_schema }}
+
+    steps:
+      - name: Check supported matrix combinations
```
bloodbare: What does this check ? Looks like we only check on PG and public ???

lferran: We don't support the following combinations of the matrix:
DATABASE="cockroach", DB_SCHEMA="custom"
DATABASE="DUMMY", DB_SCHEMA="custom"

So that was the only way I found to skip running tests on those.

.github/workflows/continuous-integration.yml
lferran: because they run on different jobs.

I could also put all in one job if you like

.github/workflows/continuous-integration.yml
bloodbare: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-excluding-configurations-from-a-matrix

.github/workflows/continuous-integration.yml
bloodbare: Maybe we could avoid installing everything on the pre-checks?

.github/workflows/continuous-integration.yml
lferran: oh nice, that must be new, because I didn't see it when I was doing it! thanks @bloodbare

bloodbare:
@lferran fallen uns tests ?  Potser el covertura nomes cal en una branca ?

lferran:
weird. seems like there is a pytest internal error when running py3.8, postgres with custom db schema:

```
  INTERNALERROR>     _remove_cached_results_from_failed_fixtures(item)
  INTERNALERROR>   File "/opt/hostedtoolcache/Python/3.8.2/x64/lib/python3.8/site-packages/pytest_rerunfailures.py", line 135, in _remove_cached_results_from_failed_fixtures
  INTERNALERROR>     result, cache_key, err = getattr(fixture_def, cached_result)
  INTERNALERROR> TypeError: cannot unpack non-iterable NoneType object
  
  ======= 1030 passed, 7 skipped, 9 warnings, 1 rerun in 217.75s (0:03:37) =======
  ##[error]Process completed with exit code 3.
```

https://docs.pytest.org/en/latest/usage.html#possible-exit-codes

lferran:
@bloodbare @vangheem @masipcat this is ready now.

I added 
```
if: matrix.database == 'postgres' && matrix.db_schema == 'public'
```
to upload covertura only once... but then the coverage decreases almost 1%.

Let me know if you prefer that I add it back despite the coverage loss.

p.s: sorry it took so long guys

lferran:
now yes! seems like pytest-rerunfailures version that we had did not support python 3.8, so when there were occasional failures, the rerun would end up in pytest internal error

PR #936, opened 2020-03-06
[5.x] Fixed validation of query params that are integers

lferran:


lferran:
@masipcat any idea why [this test](ObjectTests.test__validate_w_value_providing_schema_but_invalid_fi) is failing? it has nothing to do with my changes as far as I understand :thinking:

masipcat:
@lferran it seems the test started failed in this commit https://github.com/plone/guillotina/commit/1212c6fd0dbbe0252b1287e680e29be338f1c924

The problem is here: https://github.com/plone/guillotina/commit/1212c6fd0dbbe0252b1287e680e29be338f1c924#diff-b5c749b5b1f60f743b66487afd3aa472R21

should be `super().__init__(value, type, field_name)`

lferran:
ouch, it still fails.

vangheem:
hmmm, I don't understand how this has happened multiple times now that something was reporting that it passed...

vangheem:
@lferran I pushed fix to your branch.

masipcat:
I think there's a ghost...

lferran:
I think it was because of this https://github.com/plone/guillotina/pull/936/commits/039cda8cbaa3555950c01ff6937e6d9d6dbcf1b8

I got all tests running now without having to modify the `super().__init__(value)`

does it make sense at all?

masipcat:
Yes, because @vangheem changed the test https://github.com/plone/guillotina/pull/936/commits/9f9e8d017c9ae781ebac4fbbefc288d3f341f423 , right?

lferran:
@masipcat so you are suggesting to change back to
```
super().__init__(value, type, ...)
```
and undo nathan's changes on the test, aren't you?

PR #940, opened 2020-03-17
Use orjson instead of json/ujson

masipcat:
I'm still testing and doing stress benchmarks to ensure there is a performance gain

### DIFF
--- a/guillotina/renderers.py
+++ b/guillotina/renderers.py
@@ -8,6 +8,7 @@ from typing import cast
 from typing import Optional
 from zope.interface.interface import InterfaceClass
 
+import json
 import orjson
 
 
@@ -15,14 +16,16 @@ def guillotina_json_default(obj):
     if isinstance(obj, str):
         if type(obj) != str:  # e.g, i18n.Message()
             return str(obj)
-    if isinstance(obj, complex):
+    elif isinstance(obj, complex):
         return [obj.real, obj.imag]
-    elif isinstance(obj, datetime):
-        return obj.isoformat()
     elif isinstance(obj, type):
         return obj.__module__ + "." + obj.__name__
     elif isinstance(obj, InterfaceClass):
         return [x.__module__ + "." + x.__name__ for x in obj.__iro__]  # noqa
+    elif isinstance(obj, dict):
+        if type(obj) != dict:  # e.g. collections.OrderedDict
+            return dict(obj)
+
     try:
         iterable = iter(obj)
     except TypeError:
@@ -37,6 +40,12 @@ def guillotina_json_default(obj):
 
     # Let the base class default method raise the TypeError
     raise TypeError("Unable to serialize %r (type: %s)" % (obj, type(obj)))
+
+
+# b/w compat
+class GuillotinaJSONEncoder(json.JSONEncoder):
+    def default(self, obj):
+        return guillotina_json_default(obj)
 
 
 class Renderer:
--- a/guillotina/renderers.py
+++ b/guillotina/renderers.py
@@ -1,4 +1,3 @@
-from datetime import datetime
 from guillotina import configure
 from guillotina.interfaces import IResponse
 from guillotina.interfaces.security import PermissionSetting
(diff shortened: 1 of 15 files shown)


guillotina/request.py
```diff
@@ -239,18 +238,18 @@ def __init__(self, receive):
...
             data.extend(chunk)
 
         if n == -1:
             self._buffer = bytearray()
-            return bytes(data)
+            return data
         else:
             self._buffer = data[n:]
-            return bytes(data[:n])
+            return data[:n]
```
masipcat: I assume there is no need to convert `bytearray` to `bytes` Both types have the same interface. Also, I guess most of the time this is transformed to json isn't needed the conversion.

masipcat:
**Benchmarks**

Using molotov:
 - 60% create item. JSON body (3300 bytes)
 - 40% endpoint calls json.loads() and dumps() of a big JSON (53755 bytes)

  | json (reqps) | orjson (reqps) | diff
-- | -- | -- | --
#1 | 5064 | 5661 |  
#2 | 5423 | 5587 |  
#3 | 5107 | 5401 |  
  |   |   |  
Mean | 5198 | 5549.7 | +6.77%

Using molotov:
 - 40% create item. JSON body (3300 bytes)
 - 60% endpoint calls json.loads() and dumps() of a big JSON (53755 bytes)

  | json (reqps) | orjson (reqps) | diff
-- | -- | -- | --
#1 | 5095 | 5786 |  
#2 | 5148 | 5986 |  
#3 | 4816 | 5925 |  
  |   |   |  
Mean | 5019.7 | 5899 | +17.52%

So, we have a small improvement (+6%) for small json data and a relatively big improvement for big json data (+17%).

vangheem:
Sweet! It looks fine to me. Do we want b/w compat GuillotinaJSONEncoder still?

guillotina/renderers.py
```diff
@@ -8,33 +8,35 @@
...
+import orjson
+
+
+def guillotina_json_default(obj):
+    if isinstance(obj, str):
+        if type(obj) != str:  # e.g, i18n.Message()
+            return str(obj)
+    if isinstance(obj, complex):
+        return [obj.real, obj.imag]
+    elif isinstance(obj, datetime):
```
jordic: Is this still need with orjson?

guillotina/renderers.py
```diff
@@ -8,33 +8,35 @@
...
+        return obj.isoformat()
+    elif isinstance(obj, type):
+        return obj.__module__ + "." + obj.__name__
+    elif isinstance(obj, InterfaceClass):
+        return [x.__module__ + "." + x.__name__ for x in obj.__iro__]  # noqa
+    try:
+        iterable = iter(obj)
```
jordic: this also seems like not so optimus...

guillotina/renderers.py
masipcat: I'll remove it. I'll write a test to ensure orjson it serializes the date to the same format

guillotina/renderers.py
masipcat: I'm not sure what's the purpose of this code. It seems this is here from Guillotina 1.0.0 (sept 2016). Does anyone know if this is still needed? cc/ @bloodbare @vangheem

masipcat:
> Sweet! It looks fine to me. Do we want b/w compat GuillotinaJSONEncoder still?

I think we can add it. Is just 3 lines...

guillotina/renderers.py
masipcat: This isn't used when serializing resources because each field of the resource is serializes using the `value_serialize` (`IValueToJson` adapter). But it can be removed because orjson serializes datetime, date, and time instances to RFC 3339 format ("1970-01-01T00:00:00+00:00")

masipcat:
Ready for review!

guillotina/renderers.py
bloodbare: AFAIR its for iterables like a Set, json and ujson were not able to convert Set objects.

bloodbare:
@masipcat If you have time would u like to add a performance improvement on using orjson ? Standard perftest with json/ujson or orjson

guillotina/renderers.py
bloodbare: Also for something like dict.keys() You get and inter object not a list

PR #947, opened 2020-03-29
get_mocked_request sets the request also

bloodbare:


guillotina/contentapi.py
```diff
@@ -22,25 +22,30 @@ class ContentAPI:
     def __init__(self, db, user=RootUser("root")):
         self.db = db
         self.tm = db.get_transaction_manager()
```
masipcat: this should be None as well?

PR #948, opened 2020-04-14
Fix use of int32 sql interpolation when it should have been bigint for tid

vangheem:
This just hit us @bloodbare


vangheem:
nah, the fields were already bigint, just the sql to insert the values were using int32. So only needs changing `::int` to `::bigint`

PR #952, opened 2020-04-22
Change in ISecurityPolicy that might improve performance during traversal for views with permission guillotina.Public

masipcat:
I haven't measured the benefit of this change, but I think it might have a positive impact on applications that have a lot of request on public endpoints (like our case in @Vinissimus)

### DIFF
--- a/guillotina/security/policy.py
+++ b/guillotina/security/policy.py
@@ -62,7 +62,7 @@ class SecurityPolicy:
     @profilable
     def check_permission(self, permission, obj):
         # Always allow public attributes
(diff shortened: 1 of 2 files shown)


guillotina/security/policy.py
```diff
@@ -62,7 +62,7 @@ def invalidate_cache(self):
     @profilable
     def check_permission(self, permission, obj):
         # Always allow public attributes
-        if permission is Public:
+        if permission == Public:
```
vangheem, line 65: oh my!

PR #953, opened 2020-04-23
Fix Response object responding with 'default_content' when 'content' evaluates to False

masipcat:


guillotina/response.py
```diff
@@ -39,8 +39,10 @@ def __init__(
                 raise ValueError("body and content cannot be set at the same time")
             elif body is not None:
                 self.body = body
+            elif content is not None:
+                self.content = content
```
masipcat, line 43: Basically this solves `Response(content=[])` became `{}`

PR #956, opened 2020-04-24
[5.x] provide patch operations for json field

vangheem:


guillotina/db/interfaces.py
```diff
@@ -328,6 +328,7 @@ async def terminate(conn):
 
 class IPostgresStorage(IStorage):
     objects_table_name = Attribute("")
+    blobs_table_name = Attribute("")
```
bloodbare, line 331: Is this related to this PR?

vangheem: ahh, nah, I was in the mode of adding some mypy for some code. I can remove it.

PR #958, opened 2020-04-24
fix tags

ableeb:
some fixes for tags:
- in `services/get_endpoints`:  check if there are tags supplied in the args
- in `services/load_swagger_info`: change default to empty array so user supplied tags will be allowed


vangheem:
can you also make a pr to 5.x branch?

PR #963, opened 2020-05-05
Migration g5 to g6 fixes

lferran:


guillotina/request.py
```diff
@@ -510,9 +510,15 @@ def content(self):
...
     def content_type(self):
-        """Return raw payload stream."""
         return self.headers.get("content-type")
 
+    @reify
+    def content_length(self) -> Optional[int]:
+        try:
+            return int(self.headers["content-length"])
+        except KeyError:
+            return None
```
vangheem, line 520: Should it be 0?

masipcat: and if the users sends a string with alpha chars instead of numbers, we want to fail or return `None/0` ?

lferran: I just looked at what we had before with aiohttp https://github.com/aio-libs/aiohttp/blob/50753eace15da24a9fb9774bcfcaa1dc7650164e/aiohttp/helpers.py#L662

In this case, if user sends alpha chars, it would fail, which I think is ok no?

masipcat: Ok, yes!

guillotina/middlewares/errors.py
```diff
@@ -28,28 +33,42 @@ async def _send(msg):
...
-            "message": message,
-            "reason": error_reasons.UNKNOWN.name,
-            "details": error_reasons.UNKNOWN.details,
-            "eid": eid,
-        }
-        if app_settings.get("debug"):
-            data["traceback"] = traceback.format_exc()
-        return response.HTTPInternalServerError(content=data)
+
+def generate_error_response(
```
lferran, line 44: @masipcat @vangheem this bubbling logic was removed from 5.x to 6.x and it seems that is needed in guillotina_batch https://github.com/guillotinaweb/guillotina_batch/blob/master/guillotina_batch/api.py#L220

does this make sense?

lferran: maybe I could just re-write all of this in the guillotina_batch

masipcat: The idea was that `generate_error_response` only take care of unhandled exceptions (or exceptions that aren't a `Response`). I'd say that maybe you want to do something like this in guillotina_batch: https://github.com/plone/guillotina/commit/fd10e07f53536b77685e981d760871fe74fe8c85#diff-aab636c0eb9b37096c447b1e16abbe46R164-R175

lferran: yes! that makes sense. however, I will need to deal with unhandled exceptions there too instead of in the middleware.

thanks @masipcat

guillotina/renderers.py
```diff
@@ -17,6 +18,8 @@ def guillotina_json_default(obj):
             return str(obj)
     elif isinstance(obj, complex):
         return [obj.real, obj.imag]
+    elif isinstance(obj, datetime):
+        return obj.isoformat()
```
masipcat, line 22: @lferran what was the problem?

lferran: @masipcat this was failing

https://github.com/plone/guillotina/pull/963/files#diff-eafa441162c17595b0c457d8d08e332eR13

masipcat: I think this is expected because we removed datetime from `guillotina_json_default` because we replaced `json`/`ujson` with `orjson`, that knows how to serialize datetime objects to json (https://github.com/ijl/orjson#datetime)

I supposed that orjson would be faster because is written in rust and in that case, we wouldn't need to run the `guillotina_json_default` probably is slow (is pure-python)

PR #965, opened 2020-05-14
swagger tag values populated by provided tags, not endpoint name

ableeb:


guillotina/contrib/swagger/services.py
```diff
@@ -91,11 +91,12 @@ def get_endpoints(self, iface_conf, base_path, api_def, tags=None):
         for method in iface_conf.keys():
             if method == "endpoints":
                 for name in iface_conf["endpoints"]:
+                    for http_method in iface_conf["endpoints"][name].keys():
```
vangheem, line 94: How does this change it?

It's unclear to me what the data structure is here so we need to do:

```
tags = iface_conf["endpoints"][name][http_method][key]
```

ableeb: the tags were never actually fetched from the passed in data so we always ended up with the name of the endpoint as the tag. in the calling code https://github.com/plone/guillotina/blob/219230961065b5b56a960a819ceacd050fc5ac51/guillotina/contrib/swagger/services.py#L166, tags are not passed in, so always defaults to `None`. this change iterates over the service definition for each endpoint (`name`) and then for each http method looks for the tags.

guillotina/contrib/swagger/services.py
```diff
@@ -91,11 +91,12 @@ def get_endpoints(self, iface_conf, base_path, api_def, tags=None):
...
+                        for key in iface_conf["endpoints"][name][http_method].keys():
+                            if key in "tags":
```
masipcat, line 96: What's the type of `key`? If it's a string (and we expect the string `tags`), then this conditional should be `key == "tags"` or `key in ("tags",)`?

guillotina/contrib/swagger/services.py
```diff
@@ -91,11 +91,12 @@ def get_endpoints(self, iface_conf, base_path, api_def, tags=None):
         for method in iface_conf.keys():
             if method == "endpoints":
                 for name in iface_conf["endpoints"]:
+                    for http_method in iface_conf["endpoints"][name].keys():
+                        for key in iface_conf["endpoints"][name][http_method].keys():
```
vangheem, line 95: maybe change to 

```
if 'tags' in iface_conf["endpoints"][name][http_method]:
    tags = iface_conf["endpoints"][name][http_method]['tags']
```

guillotina/contrib/swagger/services.py
```diff
@@ -91,11 +91,11 @@ def get_endpoints(self, iface_conf, base_path, api_def, tags=None):
         for method in iface_conf.keys():
             if method == "endpoints":
                 for name in iface_conf["endpoints"]:
+                    for http_method in iface_conf["endpoints"][name].keys():
+                        if "tags" in iface_conf["endpoints"][name][http_method]:
+                            tags = iface_conf["endpoints"][name][http_method]["tags"]
```
masipcat, line 96: Does it make sense to add a `break` after we find the tags? Or we want to assign the latest value?

ableeb: yeah i think we only want the first match

svx:
🎉

PR #967, opened 2020-05-19
Spliting pg creation sentences and allowing to delete obj at cockroach

bloodbare:


guillotina/db/storages/cockroach.py
```diff
@@ -184,6 +197,16 @@ async def has_unique_constraint(self, conn):
             logger.warning("Unknown error attempting to detect constraints installed.", exc_info=True)
         return False
 
+    async def delete(self, txn, oid):
+        conn = await txn.get_connection()
+        sql = self._sql.get("CR_TRASH_PARENT_ID", self.objects_table_name)
+        deleted_id = uuid.uuid4().hex
```
vangheem, line 203: This is super weird. Why is this necessary?

The constraints should be composite for parent_id, id so you shouldn't need to change the id right?

CHANGELOG.rst
```diff
@@ -4,6 +4,12 @@ CHANGELOG
 6.0.0b4 (unreleased)
 --------------------
 
+- Allow to delete elements with the same id at cockroach
+  [bloodbare]
+
+- Split blob and objects initsenteces
```
vangheem, line 10:
```suggestion
- Split blob and objects initialization statements
```

guillotina/db/storages/cockroach.py
vangheem, line 203:
```suggestion
        # XXX since CR does not support WHERE clause in constraint, we have
        # to do this trick here to be able to mark object as deleted
        deleted_id = uuid.uuid4().hex
```

bloodbare: Any doubt ? can you review ?

vangheem:
looks like test failures though!

Go ahead and merge when fixed.

PR #969, opened 2020-05-22
Path params containing paths

dmanchon:
I wonder if accepting actual paths in the path parameter (something like this in fastapi: https://fastapi.tiangolo.com/tutorial/path-params/#path-parameters-containing-paths) could be nice to have.

So defining a service with `@api/get_path/{path:path}` with the request`@api/get_path/folder1/2/3` request will capture path as "/folder/1/2/3"

### DIFF
--- a/guillotina/contrib/swagger/services.py
+++ b/guillotina/contrib/swagger/services.py
@@ -74,7 +74,12 @@ class SwaggerDefinitionService(Service):
             route_part = route_part.strip("{}")
             if route_part not in [p["name"] for p in parameters if p.get("in") == "path"]:
                 parameters.append(
-                    {"in": "path", "name": route_part, "schema": {"type": "string"}, "required": True}
+                    {
+                        "in": "path",
+                        "name": route_part.replace(":path", ""),
+                        "schema": {"type": "string"},
+                        "required": True,
+                    }
                 )
         api_def[path.replace(":path", "") or "/"][method.lower()] = {
             "tags": swagger_conf.get("tags", []) or tags,
--- a/guillotina/contrib/swagger/services.py
+++ b/guillotina/contrib/swagger/services.py
@@ -76,7 +76,7 @@ class SwaggerDefinitionService(Service):
                 parameters.append(
                     {"in": "path", "name": route_part, "schema": {"type": "string"}, "required": True}
                 )
-        api_def[path or "/"][method.lower()] = {
+        api_def[path.replace(":path", "") or "/"][method.lower()] = {
             "tags": swagger_conf.get("tags", []) or tags,
             "parameters": parameters,
             "requestBody": request_body,
--- a/guillotina/contrib/swagger/services.py
+++ b/guillotina/contrib/swagger/services.py
@@ -33,7 +33,7 @@ class SwaggerDefinitionService(Service):
     def load_swagger_info(self, api_def, path, method, tags, service_def):
         path = path.rstrip("/")
         if path not in api_def:
-            api_def[path or "/"] = {}
+            api_def[path.replace(":path", "") or "/"] = {}
         desc = self.get_data(service_def.get("description", ""))
         swagger_conf = service_def.get("swagger", {})
         try:
--- a/guillotina/contrib/swagger/services.py
+++ b/guillotina/contrib/swagger/services.py
@@ -18,7 +18,6 @@ import pkg_resources
 
 here = os.path.dirname(os.path.realpath(__file__))
 
-
 @configure.service(
     method="GET", context=Interface, name="@swagger", permission="guillotina.swagger.View", ignore=True
 )
@@ -31,9 +30,9 @@ class SwaggerDefinitionService(Service):
         return data
 
     def load_swagger_info(self, api_def, path, method, tags, service_def):
-        path = path.rstrip("/")
+        path = path.rstrip("/").replace(":path", "")
         if path not in api_def:
-            api_def[path.replace(":path", "") or "/"] = {}
+            api_def[path or "/"] = {}
         desc = self.get_data(service_def.get("description", ""))
         swagger_conf = service_def.get("swagger", {})
         try:
@@ -73,6 +72,7 @@ class SwaggerDefinitionService(Service):
         for route_part in [r for r in service_def["route"] if r[0] == "{"]:
             route_part = route_part.strip("{}")
             if route_part not in [p["name"] for p in parameters if p.get("in") == "path"]:
+
                 parameters.append(
                     {
                         "in": "path",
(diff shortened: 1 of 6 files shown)


guillotina/traversal.py
```diff
@@ -447,9 +447,14 @@ async def real_resolve(self, request: IRequest) -> Optional[MatchInfo]:
...
 
         if not view and len(tail) > 0:
-            # we should have a view in this case because we are matching routes
-            await notify(TraversalViewMissEvent(request, tail))
-            raise HTTPNotFound(content={"reason": "object and/or route not found"})
+            # Try arbitrary "path" in the path
+            try:
+                view_name = tail[0] + "?"
+                view = query_multi_adapter((resource, request), method, name=view_name)
+            except AttributeError:
```
vangheem, line 454: I don't think this should ever raise `AttributeError`

If `view` is None here, shouldn't you be doing the `notify` and raise HTTPNotFound?

guillotina/routes.py
```diff
@@ -72,16 +80,25 @@ def __init__(self, raw_route):
...
             self.route_parts.append(RoutePart(parts[0]))
             for part in parts[1:]:
-                self.view_name += "/"
-                if URL_MATCH_RE.match(part):
-                    self.route_parts.append(VariableRoutePart(part.strip("{").strip("}")))
+                if part.endswith(":path}"):
+                    self.view_name += "?"
+                    name = part.replace(":path", "").strip("{").strip("}")
+                    self.route_parts.append(PathRoutePart(name))
+                    break
```
vangheem, line 87: I don't think the break is necessary

dmanchon: is to enforce that the "{xxx:path}" is always the last. Otherwise could be misleading if we allow the user to declare something like: "@api/{p:path}/hola" which is not valid, maybe we need to fail when we found this situation?

guillotina/traversal.py
```diff
@@ -447,9 +447,13 @@ async def real_resolve(self, request: IRequest) -> Optional[MatchInfo]:
                     )
 
         if not view and len(tail) > 0:
-            # we should have a view in this case because we are matching routes
-            await notify(TraversalViewMissEvent(request, tail))
-            raise HTTPNotFound(content={"reason": "object and/or route not found"})
+            # Try arbitrary "path" in the path
+            view_name = tail[0] + "?"
+            view = query_multi_adapter((resource, request), method, name=view_name)
+            if not view:
```
vangheem, line 453: Sorry, another tweak. Can you change `not view` to `view is None`?

Also, can you fix the `not view` on line 449?

dmanchon: ah sure!

guillotina/routes.py
vangheem: oh right, duh!

bloodbare:
> @bloodbare any objections?

I can review today later

guillotina/traversal.py
```diff
@@ -446,10 +446,14 @@ async def real_resolve(self, request: IRequest) -> Optional[MatchInfo]:
...
                     )
 
-        if not view and len(tail) > 0:
-            # we should have a view in this case because we are matching routes
-            await notify(TraversalViewMissEvent(request, tail))
-            raise HTTPNotFound(content={"reason": "object and/or route not found"})
+        if view is None and len(tail) > 0:
+            # Try arbitrary "path" in the path
+            view_name = tail[0] + "?"
+            view = query_multi_adapter((resource, request), method, name=view_name)
```
bloodbare, line 452: Why we need this here ?

dmanchon:
> I may prefer that the path tail has no parameter name, something like a boolean attribute that defines to allow any path tail and store it on a path variable. @dmanchon whats your opinion?

something like this:
```
@configure.service(context=IContainer, name='@myservice', match_tail=True, method='GET',
                   permission='guillotina.AccessContent')
async def my_service(context, request):
    ...
    filepath = request.tail 
```

vs 

```
@configure.service(context=IContainer, name='@myservice/{tail:path}', method='GET',
                   permission='guillotina.AccessContent')
async def my_service(context, request):
    filepath = request.matchdict["tail"]
```

i am ok with either solution, i find the second one slightly more readable, but maybe i am spoiled since i am used to fastapi.

guillotina/traversal.py
dmanchon: the adapter has been registered with the name `@endpoint?` instead of `@endpoint///` pattern, just a convention that we can modify to be more explicit.

guillotina/contrib/swagger/services.py
```diff
@@ -33,7 +33,7 @@ def get_data(self, data):
     def load_swagger_info(self, api_def, path, method, tags, service_def):
         path = path.rstrip("/")
         if path not in api_def:
-            api_def[path or "/"] = {}
+            api_def[path.replace(":path", "") or "/"] = {}
```
masipcat, line 36: I'm not sure if this is intentional for some reason (I'm not familiar with that part of guillotina) but looks weird to me checking the path is not in `api_def` and then using `path.replace(":path", "")` as key. Is this ok?

masipcat: Does it make sense to change this to `path = path.rstrip("/").replace(":path", "")`?

dmanchon: the real path definition unfortunately is not valid openapi :(, i am not aware if there are better ways to handle this, but producing non standard openapi could be really dangerous.

And looks like your change makes sense, i am also non familiar with openapi generation so i wanted to change as little as possible, :P

bloodbare:
> i am ok with either solution, i find the second one slightly more readable, but maybe i am spoiled since i am used to fastapi.

I just think that lead to missunderstanding that you can define the name of the variable for the path. I understand that we should be able to extend this behavior with `{VAR_NAME:int}` and `{VAR_NAME:string}`

guillotina/contrib/swagger/services.py
masipcat: > the real path definition unfortunately is not valid openapi :(, i am not aware if there are better ways to handle this, but producing non standard openapi could be really dangerous.

Yeah, I'm ok replacing the non-standard part

> And looks like your change makes sense, i am also non familiar with openapi generation so i wanted to change as little as possible, :P

Me neither :)

dmanchon:
> > i am ok with either solution, i find the second one slightly more readable, but maybe i am spoiled since i am used to fastapi.
> 
> I just think that lead to missunderstanding that you can define the name of the variable for the path. I understand that we should be able to extend this behavior with `{VAR_NAME:int}` and `{VAR_NAME:string}`

That is a very interesting idea, fastapi allow that but with a different api: https://fastapi.tiangolo.com/tutorial/path-params/
I think there is a misundertanding, you can define the name of the path var: `@endpoint/{varname: path}` then you will be able to capture `varname` in the match dictionary of the request.

PR #976, opened 2020-06-11
Be able to have optional request body

lferran:


lferran:
thoughts on this? is there already a way to do it, that I am missing?

PR #978, opened 2020-06-11
Use pendulum if it's installed

masipcat:
Sadly, I found we can't use it: https://github.com/sdispater/pendulum/issues/305


masipcat:
This can't be merged right now because of this sdispater/pendulum#305
We could consider using https://github.com/closeio/ciso8601 instead of pendulum or just add a entry in the documentation explaining how to overwrite the default deserializer with another library.

Maybe we could add a configuration entry to especify the path of the datetime parser functions, so it can be customized.

PR #981, opened 2020-06-16
Catch invalid value types for registry modifications

pfreixes:
When a new value for a specific registry is provided and the framework
complains that the type is not valid return a 412 instead of returning
a generic 500 error

### DIFF
--- a/guillotina/api/registry.py
+++ b/guillotina/api/registry.py
@@ -4,8 +4,6 @@ from guillotina.component import get_adapter
 from guillotina.event import notify
 from guillotina.events import RegistryEditedEvent
 from guillotina.exceptions import ComponentLookupError
-from guillotina.exceptions import DeserializationError
-from guillotina.exceptions import ValueDeserializationError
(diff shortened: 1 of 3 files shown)


guillotina/api/registry.py
```diff
@@ -195,12 +194,11 @@ async def __call__(self):
             return ErrorResponse(
                 "DeserializationError", "Cannot deserialize type {}".format(str(self.field)), status=412
             )
+        except ValidationError as e:
+            return ErrorResponse("ValidationError", "Invalid value type {}".format(str(e)), status=412)
 
-        try:
-            registry = await get_registry()
-            registry[self.key] = new_value
-        except (DeserializationError, ValueDeserializationError) as e:
```
pfreixes, line 202: Let me know if I broke something removing this try, except

pfreixes:
@bloodbare yeps already signed, CHANGELOG.md updated.

pfreixes:
Fixed conflict issue. PTAL @bloodbare @masipcat

PR #985, opened 2020-07-02
cleanup travis logic from tests fixtures

lferran:


lferran:
@masipcat this blocked because it has a decrease of coverage, but I'd say it's worth merging anyway right?

PR #987, opened 2020-07-08
Also allow JWT sub claim

allusa:
Allow loginid to be defined in the _sub_ (subject) JWT claim. Also maintain the private Guillotina _id_ claim.


bloodbare:
@allusa changelog needs to be changed and can you also add the sub at token generation ?

PR #988, opened 2020-07-10
Add load_catalog in test settings

masipcat:


guillotina/testing.py
```diff
@@ -51,6 +51,7 @@
     },
     "pg_connection_class": "guillotina.db.storages.pg.LightweightConnection",
     "store_json": True,
+    "load_catalog": True,
```
masipcat, line 54: Without this guillotina_elasticsearch tests doesn't work

PR #992, opened 2020-07-26
Fix sphinx-build

masipcat:
Close https://github.com/plone/guillotina/issues/984


requirements.txt
```diff
@@ -8,7 +8,7 @@ pycparser==2.18
 pycryptodome==3.6.6
 PyJWT==1.6.0
 python-dateutil==2.6.1
-PyYaml==5.1
+PyYaml>=5.1
```
masipcat, line 11: https://github.com/actions/virtual-environments/issues/1202

otherwise github actions fails

PR #994, opened 2020-07-29
Cookiecutter: fix test_install.py + test deps: unpin pytest-asyncio

masipcat:


setup.py
```diff
@@ -78,7 +78,7 @@
             "docker",
             "backoff",
             "psycopg2-binary",
-            "pytest-asyncio==0.10.0",
+            "pytest-asyncio",
```
masipcat, line 81: So we can pin the version in the application depending on the pytest version (pytest 6.0.0 requires pytest-asyncio > 0.10.0)

vangheem: Yes, it's really not a good idea to force a specific version.

masipcat:
Can you do a release please?

vangheem:
done!

masipcat:
thank you :-)

PR #996, opened 2020-07-30
Fix async test without pytest mark

masipcat:
The test was ignored by pytest.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,8 @@ CHANGELOG
 6.0.5 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Fix async test without pytest mark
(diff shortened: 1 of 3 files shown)


masipcat:
yeah... i should check the pytest warnings more often... :P

masipcat:
@vangheem don't merge this please, I want to use this branch to try to find the dependency/versions that's causing the fail

masipcat:
With pytest-asyncio 0.12.0 works fine. I replayed the jobs twice and both times passed. The problem seems to be with 0.14.0 (https://github.com/pytest-dev/pytest-asyncio/issues/178). 

I think we should keep the pin to <=13 to avoid 0.14 until is fixed. 

if this sounds good, I think we can merge this pr.

Issue #862, opened 2020-02-03
Add adapter IFileGenerateFilename


masipcat:
The idea is to be able to customize the function `generate_key()` used in guillotina_s3storage and guillotina_gcloudstorage to generate the filename of the object. 

My initial thought is:
```python
class IFileGenerateFilename():
    async def generate(file):
        ...
```

PR #1001, opened 2020-09-09
Add IFileNameGenerator adapter

Qiwn:
For GCould we can do something like this:
```python
@configure.adapter(for_=(IResource, IGCloudFileField), provides=IFileNameGenerator)
class GCloudFileNameGenerator():

    def __init__(self, context, field):
        self.context = context
        self.field = field

    def __call__(self):
        return self.context.__uuid__
```
close https://github.com/plone/guillotina/issues/862

### DIFF
--- a/guillotina/files/utils.py
+++ b/guillotina/files/utils.py
@@ -87,6 +87,4 @@ def guess_content_type(content_type, filename):
 
 
 def generate_key(context):
-    return "{}{}/{}::{}".format(
-        task_vars.container.get().id, get_content_path(context), context.__uuid__, uuid.uuid4().hex
-    )
+    return "{}/{}::{}".format(task_vars.container.get().id, context.__uuid__, uuid.uuid4().hex)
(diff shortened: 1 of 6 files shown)


guillotina/files/utils.py
```diff
@@ -87,6 +86,4 @@ def guess_content_type(content_type, filename):
 
 
 def generate_key(context):
-    return "{}{}/{}::{}".format(
-        task_vars.container.get().id, get_content_path(context), context.__uuid__, uuid.uuid4().hex
-    )
+    return "{}/{}::{}".format(task_vars.container.get().id, context.__uuid__, uuid.uuid4().hex)
```
masipcat, line 89: I'm not sure if this change can break gcloud storage / s3 storage

masipcat (approved):
Looks great!

guillotina/files/utils.py
vangheem: It shouldn't because all keys are stored persistently with the content(otherwise, moves for example wouldn't work); however, I wonder if we shouldn't change the default behavior here in case someone is depending on it?

Otherwise, this looks great!

vangheem (changes requested):

Qiwn:
ok! great! thank you!

guillotina/files/utils.py
Qiwn: :+1:

vangheem (approved):

PR #1004, opened 2020-09-15
Improve cors error

masipcat:
- TODO: add test


masipcat:
We can't return a response body on a OPTIONS request

PR #1006, opened 2020-09-22
memcached objects cache driver

lferran:


setup.py
```diff
@@ -94,6 +94,7 @@
             "sphinx-autodoc-typehints",
         ],
         "redis": ['aioredis>=1.2.0;python_version<"3.8"', 'aioredis>=1.3.0;python_version>="3.8"'],
+        "memcached": ["git+https://github.com/pfreixes/emcache"],
```
pfreixes, line 97: you can get it from pip [1] :)

[1] https://pypi.org/project/emcache/

guillotina/contrib/memcached/driver.py
```diff
@@ -0,0 +1,110 @@
...
+    async def _connect(self):
+        settings = app_settings["memcached"]
+        hosts = [emcache.MemcachedHostAddress(host, int(port)) for host, port in settings["hosts"]]
+        if len(hosts) == 0:
+            raise NoMemcachedConfigured()
+        kw = {}
+        for key in ("timeout", "max_connections"):
+            if settings.get(key) is not None:
+                kw[key] = settings[key]
+        self._client = await emcache.create_client(hosts, **kw)
```
pfreixes, line 58: Eventually, the user might be interested in using the cache cluster management interface [1]. How could we make it public?

[1] https://emcache.readthedocs.io/en/latest/cluster_managment.html

guillotina/contrib/memcached/driver.py
```diff
@@ -0,0 +1,110 @@
...
+
+    async def finalize(self):
+        if self._client is not None:
+            await self._client.close()
+        self.initialized = False
+
+    # VALUE API
+
+    async def set(self, key: str, data: str, *, expire: Optional[int] = None) -> None:
```
pfreixes, line 67: `noreply` is a good flag to consider for adding from the very beginning, is a fire and forget which increase the write throughput.

IIRC can be used in any write opreation


[1] https://emcache.readthedocs.io/en/latest/operations.html

guillotina/contrib/memcached/driver.py
```diff
@@ -0,0 +1,110 @@
...
+                        logger.error("Error initializing pubsub", exc_info=True)
+
+    @backoff.on_exception(backoff.expo, (OSError,), max_time=30, max_tries=4)
+    async def _connect(self):
+        settings = app_settings["memcached"]
+        hosts = [emcache.MemcachedHostAddress(host, int(port)) for host, port in settings["hosts"]]
+        if len(hosts) == 0:
+            raise NoMemcachedConfigured()
+        kw = {}
+        for key in ("timeout", "max_connections"):
```
pfreixes, line 55: why not support all constructor parameters from day zero [1]? They are just a handful but some of them might become really useful. 

For example `purge_unhealthy_nodes` allows you to remove temporary nodes from the pool - with a cascade of misses as a side effect - which they are considered unhealthy.


[1] https://emcache.readthedocs.io/en/latest/client.html

pfreixes: Also `connection_timeout ` is an important one.

guillotina/contrib/memcached/driver.py
lferran: User could access the `self.client` property for that: `self.client.cluser_management()`

guillotina/contrib/memcached/driver.py
lferran: after a conversation with @pfreixes, we decided not to set `noreply=True` for now.
If later we observe that is a performance problem, we can set it.

guillotina/contrib/memcached/driver.py
lferran: done all params except `cluster_events`

guillotina/contrib/memcached/driver.py
```diff
@@ -92,25 +119,29 @@ async def set(self, key: str, data: str, *, expire: Optional[int] = None) -> Non
...
-        await client.delete(key.encode())
+        with watch("delete"):
+            await client.delete(key.encode())
 
     async def delete_all(self, keys: List[str]) -> None:
         client = self._get_client()
         for key in keys:
             try:
-                await client.delete(key.encode())
+                with watch("delete_many"):
```
vangheem, line 143: These are all single deletes!
```suggestion
                with watch("delete"):
```

vangheem: Maybe you could has a watch outside the for loop to measure how long deleting all the keys inside a request take?

vangheem (changes requested):

guillotina/contrib/memcached/driver.py
lferran: good catch! yes

vangheem:
@lferran @pfreixes had pointed out that we need to record cache misses separately. See https://github.com/plone/guillotina/pull/1008 for reference.

guillotina/contrib/redis/driver.py
```diff
@@ -122,13 +122,14 @@ async def delete(self, key: str):
...
         if self._pool is None:
             raise NoRedisConfigured()
-        for key in keys:
-            try:
-                with watch("delete_many"):
+
+        with watch("delete_many"):
+            for key in keys:
+                try:
                     await self._pool.execute(b"DEL", key)
```
vangheem, line 129: metric here too then.

stubs/emcache/__init__.py
```diff
@@ -0,0 +1,46 @@
+from typing import List
```
vangheem, line 1: maybe we should contribute to emcache to provide types?

vangheem (changes requested):
looking great! Just a few comments.

stubs/emcache/__init__.py
lferran: Actually, it does provide types!
Let me remove this and check why mypy was barking

guillotina/contrib/memcached/driver.py
```diff
@@ -42,6 +61,9 @@ def __init__(self, operation: str):
 
 
 except ImportError:
+    MEMCACHED_CURRENT_CONNECTIONS = (
+        MEMCACHED_CONNECTION_POOL_CONNECTIONS
+    ) = MEMCACHED_CONNECTION_POOL_OPS = None
```
pfreixes, line 66: Maybe we can have a more explicit flag, like `_send_metrics`

guillotina/contrib/memcached/driver.py
```diff
@@ -160,3 +188,47 @@ async def flushall(self) -> None:
...
+        await asyncio.sleep(every)
+
+
+async def update_connection_pool_metrics(
+    client: emcache.Client, last_state: Optional[emcache.ConnectionPoolMetrics] = None
+) -> emcache.ConnectionPoolMetrics:
+    # Every node will have it's own label
+    metrics = client.cluster_managment().connection_pool_metrics()
+    for node, node_metrics in metrics.items():
+        MEMCACHED_CURRENT_CONNECTIONS.labels(node=node).set(node_metrics.cur_connections)
```
pfreixes, line 209: should we stringify the `node`?

guillotina/contrib/memcached/driver.py
```diff
@@ -160,3 +188,47 @@ async def flushall(self) -> None:
...
+            ],
+        }.items():
+            for type_label, metrics_attr in labels_to_attr:
+                prev_value = 0
+                if last_state is not None:
+                    try:
+                        prev_value = getattr(last_state[node], metrics_attr)
+                    except KeyError:
+                        # No metrics for that node
+                        pass
```
pfreixes, line 231: what about just a `continue` for not sending the whole counter value?

guillotina/contrib/memcached/driver.py
```diff
@@ -160,3 +188,47 @@ async def flushall(self) -> None:
...
+                ("created", "connections_created"),
+                ("created_with_error", "connections_created_with_error"),
+                ("purged", "connections_purged"),
+                ("closed", "connections_closed"),
+            ],
+            MEMCACHED_CONNECTION_POOL_OPS: [
+                ("executed", "operations_executed"),
+                ("executed_with_error", "operations_executed_with_error"),
+                ("waited", "operations_waited"),
+            ],
```
pfreixes, line 222: Should we move this to a global variable?

guillotina/contrib/memcached/driver.py
```diff
@@ -160,3 +188,47 @@ async def flushall(self) -> None:
...
+    for node, node_metrics in metrics.items():
+        MEMCACHED_CURRENT_CONNECTIONS.labels(node=node).set(node_metrics.cur_connections)
+
+        for counter, labels_to_attr in {
+            MEMCACHED_CONNECTION_POOL_CONNECTIONS: [
+                ("created", "connections_created"),
+                ("created_with_error", "connections_created_with_error"),
+                ("purged", "connections_purged"),
+                ("closed", "connections_closed"),
+            ],
```
pfreixes, line 217: IMO I would prefer to have them as independent metrics, instead of using labels for identifying each of the counters. Makes the code more implicit IMO.

vangheem (approved):

vangheem:
5.3.55 released

PR #1007, opened 2020-09-23
mailer: import 'aiosmtplib' and 'html2text' lazily

masipcat:
I'm using a custom MailerEndpoint based on aiobotocore so I don't need the aiosmtp and I don't use the html2text as well

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,9 @@ CHANGELOG
 6.0.8 (unreleased)
 ------------------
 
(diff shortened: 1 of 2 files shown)


vangheem (approved):

bloodbare:
@masipcat would not be better to define a try except of import modules. They way its implemented each time you go on that code will load the library

bloodbare (changes requested):
Try except import instead of local loading

masipcat:
I think it's only imported the first time is called, not at every function call, but I can change it

bloodbare (approved):

PR #1009, opened 2020-09-24
More metrics

vangheem:
- Record metrics on time waiting for pg locks: will help us understand how much shared connection is affecting us.
- Record hit/miss metrics on cache implementation: help us measure in-memory cache performance of root/container vs non-root objects

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,8 @@ CHANGELOG
 5.3.51 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Record metrics on time waiting for pg locks
+  [vangheem]
 
 
 5.3.50 (2020-09-23)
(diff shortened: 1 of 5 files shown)


masipcat (approved):
OMG we were discussing this with @jordic today!

vangheem:
@masipcat I know, he messaged me haha :)

vangheem:
I'm actually going to add some more metrics yet...

I'm trying to see if there is a way to isolate some reporting so we have a way of not reporting cache hit/misses on root type objects... they are screwing up the metrics since they will almost always be in cache.

bloodbare (approved):

vangheem:
@masipcat @bloodbare ready for another look.

Once we know we're done with adding metrics, I'll forward port these to master.

masipcat (approved):

PR #1015, opened 2020-10-08
Cache record size metric 

lferran:


guillotina/db/transaction.py
```diff
@@ -49,13 +49,23 @@ class ObjectResultType(TypedDict, total=False):
...
     import prometheus_client
+    from prometheus_client.utils import INF
 
     CACHE_HITS = prometheus_client.Counter(
         "guillotina_cache_ops_total",
         "Total count of ops by type of operation and the error if there was.",
         labelnames=["type", "result"],
     )
 
+    KB = 1000
```
pfreixes, line 60: 1024? xD

pfreixes (approved):

guillotina/contrib/memcached/driver.py
```diff
@@ -94,7 +94,12 @@ def __init__(self, operation: str):
 
 except ImportError:
     _SEND_METRICS = False
-    watch = metrics.watch  # type: ignore
+
+    class watch(metrics.watch):  # type: ignore
```
vangheem, line 98: oh! This might be a problem in other modules as well! Can you verify the error isn't present?

vangheem (approved):

guillotina/contrib/memcached/driver.py
lferran: I'll fix it in a separate PR

PR #1019, opened 2020-10-16
Safer keys for memcached

lferran:
## What this PR does

Hashes cache keys before sending it to memcached, to avoid having keys that aren't valid.

Pros:
 - right now, keys include db_id, and possibly the uuid of the object + container_id. The length of the key is not bounded right now. This PR fixes it in a simple way.

Cons:
 - Adds a little bit of extra cpu
 - Hashing the whole key could introduce cross-container collisions, but the probability is very low. Assuming a cache of 10Gi: 

```
prob_key_collision = 1.0 / 15**56
n_cache_keys = 10 * 2**30

prob_key_collision * n_cache_keys = ~ 1.5e-56
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ CHANGELOG
 5.3.58 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Safer keys for memcached driver [lferran]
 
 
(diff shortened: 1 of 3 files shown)


vangheem (approved):

vangheem:
5.3.58 released

PR #1020, opened 2020-10-20
Change transaction strategy 'simple'

masipcat:


guillotina/db/strategies/simple.py
```diff
@@ -10,29 +10,12 @@
...
-
-        current_tid = await self._storage.get_current_tid(self._transaction)
-        if current_tid > self._transaction._tid:
-            logger.warn(
-                f"Could not resolve conflicts in TID: {self._transaction._tid}\n"
-                f"Conflicted TID: {current_tid}"
-            )
-            return False
-
-        return True
```
masipcat, line 34: I reviewed the git history and I this was added before adding the same logic at the _store_object(), so I think that now this isn't needed. In fact, this check produces a failure when using nested transactions. The outer transaction always fails because the inner one changed the transaction id.

vangheem: This logic is a bit different than what is in store_object. Although, I don't really think simple is used at all... It is pretty "dumb" of an approach and will fail many concurrent write situations from what I can see.

masipcat: I was considering to start using it. The reason is that we have some endpoints that modify guillotina objects and do some sql inserts to another table. Without using the strategy simple we found the hard way two problems:

1. (Obviously) the asyncpg execute is done before the transaction, so it's inserted whether the guillotina transaction is commited or aborted
2. In case of TID Conflict the request is retried the 3 or 4 times so we end inserting the thing to other table 4 times

We mitigated the second point with some constraints and `INSERT INTO ... ON CONFLICT DO NOTHING` but in our case it would be great to be able to join the guillotina transaction with the real pg transaction

vangheem: Another option is to start the transaction early. We do this and it works seamlessly.

Something like:

```
    txn = get_current_transaction()
    stg = txn.storage
    if txn._db_txn is None:
        await stg.start_transaction(txn)
```

Performance on requests in these transaction blocks will be worse but it might be what you want.

masipcat: This is exactly what we want to do but we were thinking to add it globally.

> Performance on requests in these transaction blocks will be worse but it might be what you want.

We don't think that I'll be a big problem at our scale

guillotina/tests/test_postgres.py
```diff
@@ -329,34 +331,56 @@ async def test_should_not_resolve_conflict_error_with_resolve(db, dummy_guilloti
         await cleanup(aps)
 
 
-@pytest.mark.skipif(DATABASE in ("cockroachdb", "DUMMY"), reason="Cockroach not support simple...")
-async def test_should_not_resolve_conflict_error_with_simple_strat(db, dummy_guillotina):
```
masipcat, line 333: I removed this test because with this change ins't true anymore. Conflicts only happen when two transactions modify same object

guillotina/db/storages/pg.py
```diff
@@ -144,7 +144,7 @@ def _wrap_return_count(txt):
         NAIVE_UPSERT
         + """
 WHERE
-    tid = EXCLUDED.otid"""
+    {table_name}.tid = EXCLUDED.otid"""
```
masipcat, line 147: this is not used right now but while doing this PR I was testing the (non-naive) UPSERT and found that postgres was complaining about this column

guillotina/tests/test_postgres.py
```diff
@@ -329,34 +331,54 @@ async def test_should_not_resolve_conflict_error_with_resolve(db, dummy_guilloti
...
-    aps = await get_aps(db, "simple")
-    with TransactionManager(aps) as tm, await tm.begin() as txn:
-        ob1 = create_content()
-        ob2 = create_content()
-        txn.register(ob1)
-        txn.register(ob2)
+@pytest.mark.skipif(DATABASE in ("DUMMY",), reason="DUMMY not support simple...")
+async def test_should_raise_conflict_error_on_concurrent_insert(db, container_requester):
```
masipcat, line 335: For some reason the existing test was marked as `Cockroach not support simple...` but I think it works like in postgres

guillotina/tests/test_postgres.py
```diff
@@ -329,34 +331,54 @@ async def test_should_not_resolve_conflict_error_with_resolve(db, dummy_guilloti
...
+            ob = create_content(parent=container)
+            ob_bis = create_content(id=ob.id, parent=container)
+            txn1.register(ob)
+            txn2.register(ob_bis)
 
-        ob1 = await txn1.get(ob1.__uuid__)
-        ob2 = await txn2.get(ob2.__uuid__)
+            await tm.commit(txn=txn2)
+            with pytest.raises(ConflictIdOnContainer):
+                await tm.commit(txn=txn1)
```
masipcat, line 354: The conflict is catch by the pg constraint but I think it's ok

masipcat: If we want to get the ConflictError because of TID mismatch, we would need to change the UPSERT query to something like:

```diff
INSERT INTO {{table_name}}
(zoid, tid, state_size, part, resource, of, otid, parent_id, id, type, json, state)
VALUES ($1::varchar({MAX_UID_LENGTH}), $2::bigint, $3::int, $4::int, $5::boolean,
        $6::varchar({MAX_UID_LENGTH}), $7::bigint, $8::varchar({MAX_UID_LENGTH}),
        $9::text, $10::text, $11::json, $12::bytea)
- ON CONFLICT (zoid)
+ ON CONFLICT ON CONSTRAINT {constraint_name}_parent_id_id_key
DO UPDATE SET
    tid = EXCLUDED.tid,
    state_size = EXCLUDED.state_size,
    part = EXCLUDED.part,
    resource = EXCLUDED.resource,
```

vangheem (approved):

vangheem (approved):

bloodbare (approved):

PR #1021, opened 2020-10-20
Fix transaction context manager doesn't abort the txn when a exception is raised

masipcat:


guillotina/tests/test_transactions.py
```diff
@@ -31,6 +31,23 @@ async def test_tid_created_for_writes(dummy_request, event_loop):
...
+                container = await root.async_get("guillotina")
+                container.title = "changed title"
+                container.register()
+
+                raise Exception()
+
+        async with transaction(db=requester.db) as txn:
+            root = await txn.get(ROOT_ID)
+            container = await root.async_get("guillotina")
+            assert container.title == "Guillotina Container"
```
masipcat, line 48: without the fix:

```
E               AssertionError: assert 'changed title' == 'Guillotina Container'
E                 - changed title
E                 + Guillotina Container
```

guillotina/transactions.py
```diff
@@ -131,7 +131,7 @@ async def __aexit__(self, exc_type, exc, tb):
                 self.txn.deleted = {**self.previous_txn.deleted, **self.txn.deleted}
                 self.txn.added = {**self.previous_txn.added, **self.txn.added}
 
-        if self.abort_when_done:
+        if self.abort_when_done or exc:
```
masipcat, line 134: also, i guess we don't want to run execute_futures on exception, right?

vangheem: Yes, definitely!

guillotina/transactions.py
```diff
@@ -131,24 +131,27 @@ async def __aexit__(self, exc_type, exc, tb):
                 self.txn.deleted = {**self.previous_txn.deleted, **self.txn.deleted}
                 self.txn.added = {**self.previous_txn.added, **self.txn.added}
 
-        if self.abort_when_done:
+        if exc:
             await self.tm.abort(txn=self.txn)
```
masipcat, line 135: basically i changed the logic to:

if `exc` is set we only abort and restore the context vars. Otherwise, we commit/abort, we clean the previous_txn and we executed futures.

Can you confirm that this is the behavior we want?

guillotina/transactions.py
```diff
@@ -131,24 +131,27 @@ async def __aexit__(self, exc_type, exc, tb):
...
-        if self.execute_futures:
-            from guillotina.utils import execute
-
-            execute.execute_futures()
+            if self.abort_when_done:
+                await self.tm.abort(txn=self.txn)
+            else:
+                await self.tm.commit(txn=self.txn)
+
+            if self.adopt_parent_txn and self.previous_txn is not None:
```
vangheem, line 142: I think we should clear the registered changed objects no matter what. Do you agree?

vangheem (changes requested):
Just one issue about handling cleanup

guillotina/transactions.py
masipcat: I wasn't sure but it's ok :)

vangheem (approved):

bloodbare (approved):

PR #1034, opened 2020-10-22
[WIP] [Proposal] Remove TID sequence in favor of a "local" version id

masipcat:
And maybe we can add a `last_txn_timestamp` so we can still track changes globally


jordic:
so, should we add a trigger that ensures that incoming tid is always greater than stored one?

masipcat:
> so, should we add a trigger that ensures that incoming tid is always greater than stored one?

I don't see why this is needed. When the object is created we set de `serial`/`TID` to `1`. When the object is updated we increment the serial by one and during the `UPDATE` statement we check that the `serial` in pg is the `serial` before the increment (so we are updating the same object we read at the begining of the txn).

jordic:
Coll, got it should work, I just don't remember the trick on counting updated rows, seems simple than a db trigger.

PR #1036, opened 2020-11-01
Add cors wrapper on conflict response

bloodbare:


masipcat (approved):
Looks good to me :) Maybe in a future (G6.x / G7) we could move the function apply_cors to a ASGI middleware

PR #1043, opened 2020-11-11
Improve delete all for the memcache driver

pfreixes:
- Do not execute internally the delete all when len of the keys is 0
- Observe, if metrics available, the size of the list of keys

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -81,7 +81,7 @@ setup(
             "pytest-asyncio==0.10.0",
             "pytest-cov>=2.0.0,<=2.9.0",
             "coverage>=4.0.3",
-            "pytest-docker-fixtures",
+            "pytest-docker-fixtures==1.3.11",
             "pytest-rerunfailures<10.0.0",
             "openapi-spec-validator",
             "asyncmock",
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -81,7 +81,7 @@
             "pytest-asyncio==0.10.0",
             "pytest-cov>=2.0.0,<=2.9.0",
             "coverage>=4.0.3",
-            "pytest-docker-fixtures",
+            "pytest-docker-fixtures==1.3.11",
```
vangheem, line 84: Is this pin necessary?

vangheem (changes requested):

setup.py
pfreixes: this is the version that was released with support for the memcached fixture

vangheem: Can we do `>=`?

vangheem:
Also, just a real simple flake issue:

```
guillotina/tests/memcached/test_memcached_driver.py:137:100: E231 missing whitespace after ','
```

pfreixes:
@vangheem Feedback applied, PTAL

vangheem (approved):

vangheem:
5.3.61 released.

PR #1048, opened 2020-11-25
[5.x] Fix not recording pg root metrics properly

vangheem:
Objects that come from an external cache system are `dict` type; however, objects that are from in-memory and misses are `asyncpg.Record` type.

### DIFF
--- a/guillotina/db/transaction.py
+++ b/guillotina/db/transaction.py
@@ -31,6 +31,7 @@ from typing_extensions import TypedDict
 from zope.interface import implementer
 
 import asyncio
+import asyncpg
 import logging
 import sys
 import time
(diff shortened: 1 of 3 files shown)


guillotina/db/transaction.py
```diff
@@ -359,6 +360,7 @@ async def refresh(self, ob):
         additional_keys=[lambda item: {"container": item["parent_id"], "id": item["id"]}],
     )
     async def _get(self, oid):
+        print("loading")
```
pfreixes, line 363: leftovers xD

pfreixes (approved):

masipcat (approved):

Issue #1026, opened 2020-10-22
Store JSON objects instead of pickle objects


bloodbare:
Right now base object stores pickles which are fast but does not allow to do complex SQL queries and provide introspection with other languages than python.

PR #1051, opened 2020-11-27
[PoC] Store JSON objects instead of pickle objects

masipcat:
https://github.com/plone/guillotina/issues/1026

### DIFF
--- a/guillotina/db/serializers/json/value_deserializers.py
+++ b/guillotina/db/serializers/json/value_deserializers.py
@@ -1,0 +1,105 @@
+from .interfaces import IStorageDeserializer
+from dateutil.parser import parse
+from guillotina import configure
+from guillotina.interfaces.security import PermissionSetting
+from guillotina.security.securitymap import SecurityMap
+from guillotina.utils import resolve_dotted_name
+from zope.interface import Interface
+from zope.interface.declarations import Implements
+
+import base64
+import pickle
+
+
+@configure.adapter(
+    for_=Interface, provides=IStorageDeserializer, name="$.AnyObjectDict",
+)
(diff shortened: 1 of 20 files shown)


masipcat:
Almost everything works. Only need to fix tests that manipulate files

masipcat:
All tests green! But there are some things to consider/discuss:

1. I think the current implementation is vulnerable. If someone sends a dict like `{"__class__": ...}` on a JSONField, we'll resolve the dotted name and create an instance of the `__class__` when reading the object from db. The first thing that comes to my mind to mitigate this is appending a secret key like this  `__class__y2o930874h5f8764598`. This requires each guillotina installation to generate this key and put in in the config. What do you think? Any other idea on how to solve it?

2. orjson is strict and doesn't serialize keys that aren't strings and I found some objects like Blob and BucketList that have dicts with numeric keys. I solved these cases doing a pickle of the object :P This is something that the developer of the application can solve for their own types, but needs to implement custom `IStorageSerializer`/`Deserializer` and might be a little inconvenient.

guillotina/db/serializers/json/value_deserializers.py
```diff
@@ -0,0 +1,128 @@
...
+    for_=Interface, provides=IStorageDeserializer, name="zope.interface.Provides",
+)
+class ProvidesDeserializer:
+    def __init__(self, data):
+        self.data = data
+
+    def __call__(self):
+        # from zope.interface import Provides  # type: ignore
+        # obj = Provides(*[resolve_dotted_name(iface) for iface in self.data["__ifaces__"]])
+        # return obj
```
masipcat, line 100: Does anyone know how to make it work?

PR #1053, opened 2020-11-30
Update Transaction.contains, Transaction.tpc_commit

Qiwn:
Update Transaction.contains, it also checks in the deleted dict
Transaction.tpc_commit: process deleted items before added items (When we want to remove an existing obj, and add a new obj with the same id)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,8 @@ CHANGELOG
 5.3.63 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Update Transaction.contains
+  [qiwn]
 
 
 5.3.62 (2020-11-25)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ CHANGELOG
 5.3.63 (unreleased)
 -------------------
 
-- Update Transaction.contains
+- Update Transaction.contains, Transaction.tpc_commit
   [qiwn]
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,10 @@ CHANGELOG
 5.3.63 (unreleased)
 -------------------
 
-- Update Transaction.contains, Transaction.tpc_commit
+- Check deleted list in Transaction.contains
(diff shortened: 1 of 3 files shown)


guillotina/db/transaction.py
```diff
@@ -556,7 +556,13 @@ async def get_children(self, parent, keys):
 
     @profilable
     async def contains(self, oid: str, key: str) -> bool:
-        return await self._manager._storage.has_key(self, oid, key)  # noqa
+        if not await self._manager._storage.has_key(self, oid, key):  # noqa
+            return False
+
+        for obj in self.deleted.values():
+            if obj.id == key:
```
vangheem, line 563: I think you also need to make sure the object's parent uid equals `oid` right?

guillotina/db/transaction.py
```diff
@@ -556,7 +556,13 @@ async def get_children(self, parent, keys):
 
     @profilable
     async def contains(self, oid: str, key: str) -> bool:
-        return await self._manager._storage.has_key(self, oid, key)  # noqa
+        if not await self._manager._storage.has_key(self, oid, key):  # noqa
+            return False
+
+        for obj in self.deleted.values():
```
vangheem, line 562: Should this check happen before the `self._manager._storage.has_key` line?

vangheem (changes requested):

guillotina/db/transaction.py
Qiwn: right!

CHANGELOG.rst
```diff
@@ -4,7 +4,8 @@ CHANGELOG
 5.3.63 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Update Transaction.contains, Transaction.tpc_commit
```
vangheem, line 7: Can you describe what you're updating here? That contains is now checking the current app txn and the commit ordering has changed?

vangheem (changes requested):

vangheem (approved):

vangheem:
5.3.63 released

Qiwn:
thank you!

PR #1057, opened 2020-12-03
Docs2020

bloodbare:
- metadata endpoint
- suggestion endpoint

### DIFF
--- a/docs/source/contrib/email_validation.md
+++ b/docs/source/contrib/email_validation.md
@@ -1,0 +1,91 @@
+# Email validation
+
+`guillotina` provides out of the box an email validation system to validate actions with a mail with a link that frontend needs to negotatiate with the API.
+
+Its scalable on defining different actions that need email validation. By default Guillotina provides Register user and Reset password actions email validation.
+
+This addon requires Installing addon `email_validation` to register the persistent settings on the container.
+
+## Configuration
+
+### Add a new action
+
+You need to define on the config.yaml or package configuration::
+
+```yaml
+auth_validation_tasks:
+    reset_password:
+        schema:
+            title: Reset password validation information
+            required:
+            - password
+            type: object
+            properties:
+                password:
+                    type: string
+                    widget: password
+                    minLength: 6
+        executor: guillotina.contrib.email_validation.reset_password
+```
+
+This configuration defined the action `reset_password`, which is the schema that will need to be validated on 2nd fase flow and who is the executor of the action after validation.
+The executor should be a module with an async `run` function.
+
+### Container configuration
+
+You can configure at interface `guillotina.contrib.email_validation.interfaces.IValidationSettings`:
+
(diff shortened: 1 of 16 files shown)


docs/source/contrib/email_validation.md
```diff
@@ -0,0 +1,91 @@
+# Email validation
+
+`guillotina` provides out of the box an email validation system to validate actions with a mail with a link that frontend needs to negotatiate with the API.
+
+Its scalable on defining different actions that need email validation. By default Guillotina provides Register user and Reset password actions email validation.
+
+This addon requires Installing addon `email_validation` to register the persistent settings on the container.
```
svx, line 7:
```suggestion
This addon requires Installing the `email_validation` addon to register the persistent settings on the container.
```

docs/source/contrib/email_validation.md
```diff
@@ -0,0 +1,91 @@
...
+            type: object
+            properties:
+                password:
+                    type: string
+                    widget: password
+                    minLength: 6
+        executor: guillotina.contrib.email_validation.reset_password
+```
+
+This configuration defined the action `reset_password`, which is the schema that will need to be validated on 2nd fase flow and who is the executor of the action after validation.
```
svx, line 31:
```suggestion
This configuration defines the action `reset_password`, which is the schema that will need to be validated on 2nd fase flow and who is the executor of the action after validation.
```

docs/source/contrib/email_validation.md
```diff
@@ -0,0 +1,91 @@
...
+            task_description="Reset password",  # This is the email summary
+            task_id="reset_password",  # This is the action id
+            context_description=self.context.title,  # Context for the mail
+            redirect_url=redirect_url,  # After validation redirect url
+            data={}  # Base data for the 1rst step validation
+        )
+        ...
+```
+
+Once the mail is sent and the user clicks on the link frontend can:
```
svx, line 71:
```suggestion
Once the mail is sent and the user clicks on the link the frontend can:
```

svx (commented):
Two tiny things 😄 

BTW, looks like I will have some time for the G-Docs from January on 😄 !

guillotina/api/suggestion.py
```diff
@@ -0,0 +1,58 @@
...
+                }
+            }
+        }
+    },
+)
+async def suggestion_get(context, request):
+    query = request.query.copy()
+    search = query_utility(ICatalogUtility)
+    if search is None:
+        return {}
```
masipcat, line 35: I think it's confusing to return empty result when ICatalogUtility is None. You don't know if it's because the query doesn't match any result or you forgot to configure the catalog. I'd say the same for GET/POST @search endpoints.

What do you think?

bloodbare: Agree

masipcat (approved):

PR #1058, opened 2020-12-04
Fix metrics missing variable

masipcat:


guillotina/db/storages/pg.py
```diff
@@ -67,7 +67,21 @@ def __init__(self, lock: asyncio.Lock, operation: str):
 
 
 except ImportError:
-    watch = metrics.watch  # type: ignore
+
+    class watch:
```
bloodbare, line 71: This can also be metics.watch because Counter will be None, no ?

guillotina/metrics.py
```diff
@@ -78,11 +78,12 @@ def __init__(
     async def __aenter__(self) -> None:
         start = time.time()
         await self.lock.acquire()
-        finished = time.time()
-        if len(self.labels) > 0:
-            self.histogram.labels(**self.labels).observe(finished - start)
-        else:
-            self.histogram.observe(finished - start)
+        if self.histogram:
```
vangheem, line 81:
```suggestion
        if self.histogram is not None:
```

vangheem (changes requested):
maybe not none check instead?

vangheem (approved):

bloodbare (approved):

PR #1068, opened 2020-12-21
Fixing dependenciy

bloodbare:


masipcat (commented):
I think it needs to be added to setup.py as well

masipcat (approved):

PR #1072, opened 2020-12-23
Fixing addable types

bloodbare:


guillotina/api/content.py
```diff
@@ -716,9 +716,7 @@ async def addable_types(context, request):
     constrains = IConstrainTypes(context, None)
     types = constrains and constrains.get_allowed_types()
     if types is None:
-        types = []
-        for type_name in FACTORY_CACHE:
-            types.append(type_name)
+        types = [type_name for type_name in FACTORY_CACHE if constrains.is_globally_allowed(type_name)]
     app_settings["container_types"]
```
masipcat, line 720: Not related to this PR but: what's the purpose of this `app_settings["container_types"]` ? Checking the key exists ans failing otherwise?

masipcat (approved):

guillotina/api/content.py
bloodbare: I think its and old leak of code. Cleaning it

PR #1073, opened 2021-01-03
Fixing keyword not search and aggregation endpoint

bloodbare:


masipcat (commented):
I'm not sure, maybe we need to update this line as well? https://github.com/plone/guillotina/blob/01fc3fe3dbed0e4f1b85b9ae2a89be6df6755026/guillotina/contrib/catalog/pg.py#L206

bloodbare:
> I'm not sure, maybe we need to update this line as well?
> 
> https://github.com/plone/guillotina/blob/01fc3fe3dbed0e4f1b85b9ae2a89be6df6755026/guillotina/contrib/catalog/pg.py#L206

The PR is to fix the __not on keywords. Right now is not working because ? and ?| are not working. Which use case are you referencing?

masipcat:
I was thinking:

```python
if operator == "?" and isinstance(result, list):
    operator = "?|"
elif operator == "NOT ?" and isinstance(result, list):
    operator = "NOT ?|"
```

masipcat (approved):

PR #1075, opened 2021-01-07
Filename on dict download

bloodbare:


guillotina/api/files.py
```diff
@@ -164,8 +172,9 @@ async def __call__(self):
         # We need to get the upload as async IO and look for an adapter
         # for the field to save there by chunks
         kwargs = {}
-        if "filename" in self.request.matchdict:
-            kwargs["filename"] = self.request.matchdict["filename"]
+        filename = self.request.query.get("filename")
+        if filename is not None:
+            kwargs["filename"] = filename
```
masipcat, line 177: if we use `file_key` as a fallback for `self.request.query.get("filename")`, then it will be backwards compatible?
```suggestion
        filename = self.request.query.get("filename")
        if filename is not None:
            kwargs["filename"] = filename
        elif "file_key" in self.request.matchdict:
            kwargs["filename"] = self.request.matchdict["file_key"]
```

bloodbare: If you query /@download/field/dict_key the downloaded file will have dict_key as filename

masipcat: yes, but this is the current behavior, right? I mean, if we want to stop doing this it's fine, but if we only want to be able to customize the filename, maybe we can keep the previous behavior

bloodbare: Right now you can not download a file with the origin filename if its a Dict cloud file

masipcat (approved):

PR #1077, opened 2021-01-09
Fixing a bug in TUS

bloodbare:


guillotina/api/service.py
```diff
@@ -250,7 +250,7 @@ async def prepare(self):
             await self.behavior.load()
 
         if IDict.providedBy(field) and ICloudFileField.providedBy(field.value_type):
-            key = self.request.matchdict["file_key"]
+            key = self.request.matchdict.get("file_key")
```
masipcat, line 253: what is fixing this change?

bloodbare: In case you request OPTIONS of /@tusupload/files_dict without the file key it returns a 500, which should return a 400

masipcat (approved):

PR #1081, opened 2021-01-11
Fix bug passing emcache settings

lferran:


pfreixes (approved):

bloodbare:
@pfreixes @lferran would be great that 5.x is not in active development and features and fixes are applied to 6. Guillotina 7 plans are already on the way and would be great to join efforts!

lferran:
> @pfreixes @lferran would be great that 5.x is not in active development and features and fixes are applied to 6. Guillotina 7 plans are already on the way and would be great to join efforts!

@bloodbare totally agree! I can port-forward the `contrib.memcached` to master. This is a bugfix though

vangheem:
5.3.65 released

PR #1085, opened 2021-01-13
Fix: prevent caching objects bigger than max_cache_record_size 

lferran:
Seems that we are checking this restriction [during the transaction](https://github.com/plone/guillotina/blob/5.x/guillotina/db/transaction.py#L128), but not when syncronizing through the pubsub [at transaction close time](https://github.com/plone/guillotina/blob/5.x/guillotina/contrib/cache/strategy.py#L144). 

Therefore, we are storing in memory and sending to redis/memcached objects that are over the size limit.

Not 100% sure if that's the best way to fix it though, as it could also be done in `BasicCache.fill_cache`.

Once this is correctly fix, I will port to `master`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ CHANGELOG
 5.3.66 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Fix: prevent caching large objects on fill_cache [lferran]
 
 
 5.3.65 (2021-01-11)
--- a/guillotina/contrib/cache/strategy.py
+++ b/guillotina/contrib/cache/strategy.py
@@ -81,8 +81,7 @@ class BasicCache(BaseCache):
     async def store_object(self, obj, pickled):
         if len(pickled) > self.max_cache_record_size:
             # Do not store objects that exceed the record size
-            # return
-            pass
+            return
 
         if len(self._stored_objects) < self.max_publish_objects:
             self._stored_objects.append((obj, pickled))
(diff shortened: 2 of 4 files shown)


vangheem:
It makes sense to put it where you have it!

masipcat:
Good catch! 

Just one thing: I think with your changes the [check_state_size](https://github.com/plone/guillotina/blob/dfa7e9e77a16722f630216e9ac5f055a1ff34350/guillotina/db/transaction.py#L98) doesn't make sense anymore (it's not used). Maybe we can make `max_cache_record_size` nullable in case someone doesn't want the limit

lferran:
> Good catch!
> 
> Just one thing: I think with your changes the [check_state_size](https://github.com/plone/guillotina/blob/dfa7e9e77a16722f630216e9ac5f055a1ff34350/guillotina/db/transaction.py#L98) doesn't make sense anymore (it's not used). Maybe we can make `max_cache_record_size` nullable in case someone doesn't want the limit

Yes! agree on your idea. I'd prefer doing it in a separate PR though as a chore (maybe someone is actually using `check_state_size`).

lferran:
ready for review

(don't know why codecov step is failing :cry:)

masipcat:
> ready for review
> 
> (don't know why codecov step is failing cry)

maybe it's because the added/deleted empty lines in `guillotina/db/transaction.py`

masipcat (approved):

vangheem:
5.3.66 released

Issue #1069, opened 2020-12-22
DBUsers Patch Group deletes group from user


bloodbare:
If you  patch a group from dbusers with only user_roles, the group from the user objects is deleted. 

PATCH: ```{"user_roles":["guillotina.Reader"]}```

May be related with https://github.com/plone/guillotina/blob/master/guillotina/contrib/dbusers/subscribers.py#L68

PR #1093, opened 2021-01-17
Dbusers improvements

jordic:
- Fixes bug https://github.com/plone/guillotina/issues/1069
- Be able to change a user password by patching it.


CHANGELOG.rst
```diff
@@ -5,6 +5,8 @@ CHANGELOG
 ------------------
 
 - Fix: prevent caching large objects on fill_cache [lferran]
+- Fix: patch a role on dbusers:groups don't remove users from them
```
bloodbare, line 8: Could u add your username?

guillotina/tests/dbusers/test_manage_users.py
```diff
@@ -114,6 +114,19 @@ async def test_patch_user_data(dbusers_requester, user_data):
         assert resp["email"] == "foobar2@foo.com"
         assert "guillotina.Manager" in resp["roles"]
 
+        # Change password
+        data = {"password": "asdf"}
+        resp, status_code = await requester("PATCH", "/db/guillotina/@users/foobar", data=json.dumps(data))
```
bloodbare, line 119: There is already the @reset-password endpoint to set the password of your user, this feature would allow root user to set user password. Is this correct ?

masipcat: IMO it's ok that root can set users' password. Also, that a user can change his own password while is logged (without the need to send an email), but in that case maybe we want to check the user know the current password to set a new one.

CHANGELOG.rst
masipcat: and move your lines above the existing ones :P (so it matches the git history)

masipcat (approved):

guillotina/tests/dbusers/test_manage_users.py
jordic: Screenshot from 2021-01-17 12-18-47

jordic: this is a facility to allow root user to change passwords. A user can change it's password if it's logged, (though) this is a side effect, I think it's better to just go with tthis, I don't agree on complicating the endpoint checkin the current user password, at the most, disallow using this feature if the user it's not manager.

jordic: BTW, the feature also fixes the issue around hashing passwords on PATCH operations.

guillotina/contrib/dbusers/users.py
```diff
@@ -40,3 +46,51 @@ async def get_user(self, token: typing.Dict) -> typing.Optional[IPrincipal]:
...
+
+    async def get_user(self, token: typing.Dict) -> typing.Optional[IPrincipal]:
+        try:
+            container = get_current_container()
+            users = await container.async_get("users")
+        except (AttributeError, KeyError, ContainerNotFound):
+            return None
+
+        catalog = query_utility(ICatalogUtility)
+        if catalog.__class__ != PGSearchUtility:
```
bloodbare, line 62: Can it be isinstance ?

bloodbare (changes requested):

bloodbare (approved):

masipcat (approved):

PR #1097, opened 2021-01-29
Adding local roles on @search scope

bloodbare:


masipcat (approved):

bloodbare:
@masipcat you should also apply it to guillotina elasticsearch, as it does not check local roles of the user

guillotina/contrib/catalog/pg.py
```diff
@@ -487,6 +488,7 @@ def get_default_where_clauses(self, container: IContainer) -> typing.List[str]:
         users.extend(principal.groups)
         roles_dict = policy.global_principal_roles(principal.id, principal.groups)
         roles.extend([key for key, value in roles_dict.items() if value])
+        roles = get_roles_principal(container)
```
vangheem, line 491: You are overriding the `roles` value used above. Is that on purpose?

vangheem: Should it be an extension?

bloodbare: ops I forgot to delete it, get_roles_principal already gets the global ones, so its not an extend

vangheem: ahh right, that makes sense.

bloodbare: Fixed at https://github.com/plone/guillotina/commit/bd1468066189c8fe95b52a1f0002070b54aaeafd (sorry I did not do a PR)

PR #1098, opened 2021-02-02
add max_content_depth limit

lferran:
This PR does:

- Adds a configurable limit (`max_content_depth`) on height of the objects tree.

The motivation behind it is because there are some recursive functions in guillotina ( see [cached_principal_perimssion](https://github.com/plone/guillotina/blob/master/guillotina/security/policy.py#L140), for instance) where if we don't put a limit, we could run into python throwing `RecursionError`. We've found this at Onna.

I'll forward-port once we agree on the solution and it's merged.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,8 +4,7 @@ CHANGELOG
 5.3.67 (unreleased)
 -------------------
 
-- Nothing changed yet.
-
+- Add max_content_depth [lferran]
 
 5.3.66 (2021-01-13)
 -------------------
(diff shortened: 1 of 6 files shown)


vangheem:
@bloodbare @masipcat  how do you feel about this?

masipcat:
It's ok for me to add it, but I think it's still possible to bypass this restriction when moving folder

vangheem (approved):

bloodbare (approved):
I agree that perfect solution would be with move/copy restriction. As it implementation would require walking to get depth of children I think we should merge this right now.

PR #1099, opened 2021-02-03
improve memcached metrics probe

lferran:
This PR improves in two ways:
- getting metrics no need to be async (as there is no IO ops)
- no need to deepcopy the metrics dict every time, as [emcache already returns a copy every time](https://github.com/pfreixes/emcache/blob/master/emcache/connection_pool.py#L348)

Will forward to master later, once merged


vangheem (approved):

vangheem:
5.3.67 released

PR #1101, opened 2021-02-08
Fixing field permission on create content

bloodbare:


guillotina/api/content.py
```diff
@@ -224,13 +231,6 @@ async def __call__(self):
 
         await deserializer(data, validate_all=True, create=True)
 
-        # Local Roles assign owner as the creator user
-        get_owner = IGetOwner(obj)
```
vangheem, line 228: This seems like a rather big change to not have owner assignment anymore.

guillotina/api/content.py
```diff
@@ -212,6 +212,13 @@ async def __call__(self):
         for behavior in behaviors or ():
             obj.add_behavior(behavior)
 
+        # Local Roles assign owner as the creator user
```
vangheem, line 215: Why move this up?

guillotina/_settings.py
```diff
@@ -97,5 +97,12 @@
     "valid_id_characters": string.digits + string.ascii_lowercase + ".-_@$^()+ =",
     "load_catalog": True,
     "catalog_max_results": 50,
+    "managers_roles": {
```
vangheem, line 100: can we rename `managers_roles` -> `manager_roles`?

vangheem: Good idea!

vangheem (changes requested):

guillotina/api/content.py
bloodbare: If owner is not set before deserialization field permissions are not available as guillotina.Owner on the specific object

bloodbare (commented):
the group is called Managers

guillotina/_settings.py
bloodbare: the group is called managers

vangheem: But they are `roles` a "Manager" will have. Dunno, just sounds funny to me. You can leave it.

vangheem (approved):

masipcat (approved):

lferran:
@bloodbare I'm trying to fix memcached tests. You may want to merge this one on the meantime: https://github.com/plone/guillotina/pull/1102

lferran:
@bloodbare instead of commenting tests, can you merge this one? https://github.com/plone/guillotina/pull/1103

Should be ready now

PR #1103, opened 2021-02-09
Fix memcached tests

lferran:
For some reason `pytest-docker-fixtures` is not able to start the memcached image anymore when running the github actions job (locally it works though).

Adding server startup in a separate action step made the fix.

### DIFF
--- a/guillotina/tests/fixtures.py
+++ b/guillotina/tests/fixtures.py
@@ -638,9 +638,19 @@ def redis_container():
 
 @pytest.fixture(scope="session")
 def memcached_container():
-    host, port = os.environ["MEMCACHED"].split(":")
-    annotations["memcached"] = (host, port)
-    yield host, port  # provide the fixture value
+    if os.environ.get("MEMCACHED"):
+        host, port = os.environ["MEMCACHED"].split(":")
+        annotations["memcached"] = (host, port)
+        yield host, port
+
+    else:
+        import pytest_docker_fixtures
+
+        host, port = pytest_docker_fixtures.memcached_image.run()
+        annotations["memcached"] = (host, port)
+        yield host, port
+        pytest_docker_fixtures.memcached_image.stop()
+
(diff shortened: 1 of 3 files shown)


guillotina/tests/fixtures.py
```diff
@@ -638,12 +638,9 @@ def redis_container():
 
 @pytest.fixture(scope="session")
 def memcached_container():
-    import pytest_docker_fixtures
-
-    host, port = pytest_docker_fixtures.memcached_image.run()
+    host, port = os.environ["MEMCACHED"].split(":")
```
bloodbare, line 641: If env var is not defined what happens?

bloodbare (changes requested):

guillotina/tests/fixtures.py
lferran: yup, right. Added back using pytest docker fixtures if env var is not configured

guillotina/tests/fixtures.py
```diff
@@ -638,12 +638,19 @@ def redis_container():
...
+        annotations["memcached"] = (host, port)
+        yield host, port
+
+    else:
+        import pytest_docker_fixtures
+
+        host, port = pytest_docker_fixtures.memcached_image.run()
+        annotations["memcached"] = (host, port)
+        yield host, port
+        pytest_docker_fixtures.memcached_image.stop()
```
masipcat, line 652: https://github.com/guillotinaweb/pytest-docker-fixtures/commit/209e1311ae7d1f6d3d60ad7cf6be96065df8651b#diff-7b90426c13035fba59af80ec62fb738e3cf4361fe5d5d0ee2ddbdc37b5300dd7R127-R135

bloodbare (approved):

bloodbare:
@lferran I merged because I need to merge that other branch, please finish what @masipcat said

PR #1105, opened 2021-02-24
Fix Behavior.__getattr__() returns a new instance of default value each time is called

masipcat:
This is a continuation of https://github.com/plone/guillotina/pull/395 (https://github.com/plone/guillotina/issues/393)


bloodbare (changes requested):
It may set values on accessing default values on the behavior on PATCH/POST views

masipcat:
-> https://github.com/plone/guillotina/pull/1108

PR #1108, opened 2021-03-03
Write missing_value to object on deserialization

masipcat:
This change is to avoid chaning the current behavior of `default` (https://github.com/plone/guillotina/pull/1105#pullrequestreview-597746430)


guillotina/json/deserialize_content.py
```diff
@@ -117,11 +117,14 @@ async def set_schema(
...
                 data_value = data[name] if name in data else None
                 found = True if name in data else False
-            if found:
 
-                if not self.check_permission(write_permissions.get(name)):
+            if found or field.missing_value is not None:
+                if found and not self.check_permission(write_permissions.get(name)):
                     raise Unauthorized("Write permission not allowed")
 
+                if field.missing_value is not None:
```
bloodbare, line 125: This will not set the value always to the missing_value if its defined?

bloodbare (changes requested):

guillotina/json/deserialize_content.py
masipcat: ups, yes

masipcat, line 125:
```suggestion
                if not found and field.missing_value is not None:
```

guillotina/json/deserialize_content.py
```diff
@@ -117,11 +117,14 @@ async def set_schema(
...
                 found = True if name in data else False
-            if found:
 
-                if not self.check_permission(write_permissions.get(name)):
+            if found or field.missing_value is not None:
+                if found and not self.check_permission(write_permissions.get(name)):
                     raise Unauthorized("Write permission not allowed")
 
+                if not found and field.missing_value is not None:
+                    data_value = deepcopy(field.missing_value)
```
bloodbare, line 126: This will happen for each patch for each missing value no matter if the field has already value. I would do it later.

guillotina/json/deserialize_content.py
```diff
@@ -117,11 +117,14 @@ async def set_schema(
...
+
```
masipcat, line 127:
```suggestion
            # Only set missing_value when it's not present in 'data' and we are validating_all fields
            must_set_mv = not found and validate_all and field.missing_value is not None

            if found or must_set_mv:
                if found and not self.check_permission(write_permissions.get(name)):
                    raise Unauthorized("Write permission not allowed")

                if must_set_mv:
                    data_value = deepcopy(field.missing_value)

```

guillotina/json/deserialize_content.py
masipcat: What do you think of this: https://github.com/plone/guillotina/pull/1108/files/60b4423d47c92d5955ea7f5140213087cef9564f#r586610866

guillotina/json/deserialize_content.py
bloodbare: Its better you do a commit than yourself a suggestion its difficult to comment on specific line. This change may involve a missing value on a field that user does not have permission to set which will return on a 401.

masipcat: already commited ;-)

>  This change may involve a missing value on a field that user does not have permission to set which will return on a 401.

We only check the permission if the field is found in the data.

bloodbare (approved):

PR #1115, opened 2021-04-07
Enabling option to wait for futures to finish

bloodbare:


guillotina/traversal.py
```diff
@@ -277,9 +283,12 @@ async def handler(self, request):
...
 
         if not request._view_error:
-            request.execute_futures()
+            task = request.execute_futures()
         else:
-            request.execute_futures("failure")
+            task = request.execute_futures("failure")
+
+        if self.wait(request):
+            await asyncio.wait_for(task, app_settings.get("max_wait_futures"))
```
vangheem, line 291: You want the task to be cancelled if it doesn't finish in time?

if not, you should use `asyncio.wait` instead.

vangheem (approved):

guillotina/traversal.py
bloodbare: yups, that is max_wait_futures settings . The idea is, if you want to wait, you want to define a max to wait. It may be conflictive if it fails and commit is already done.

vangheem: Yeah, then don't use wait_for

guillotina/traversal.py
```diff
@@ -287,8 +290,11 @@ async def handler(self, request):
         else:
             task = request.execute_futures("failure")
 
-        if self.wait(request):
-            await asyncio.wait_for(task, app_settings.get("max_wait_futures"))
+        wait = self.wait(request)
```
bloodbare, line 293: @vangheem what do you think :)

vangheem: no way IMO

All the APIs also say "after_request"

vangheem: eh, up to you. Seems a bit crazy IMO.

bloodbare: Ok lets wait not for it :)

vangheem (approved):

PR #1119, opened 2021-04-16
return 401 if NoPermissionToAdd is raised in defaultPOST

nilbacardit26:
Currently, I am getting 500 when NoPermissionToAdd is raised in the defaultPOST method when create_content_in_container is called. This happens when I am configuring a contentype with a custom add_permission. Raise HTTPUnauthorized instead

### DIFF
--- a/guillotina/api/content.py
+++ b/guillotina/api/content.py
@@ -24,7 +24,6 @@ from guillotina.events import ObjectPermissionsViewEvent
 from guillotina.events import ObjectRemovedEvent
 from guillotina.events import ObjectVisitedEvent
 from guillotina.exceptions import ComponentLookupError
-from guillotina.exceptions import NoPermissionToAdd
 from guillotina.exceptions import PreconditionFailed
 from guillotina.i18n import default_message_factory as _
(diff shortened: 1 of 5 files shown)


guillotina/api/content.py
```diff
@@ -206,6 +207,8 @@ async def __call__(self):
             obj = await create_content_in_container(
                 self.context, type_, new_id, check_constraints=True, **options
             )
+        except NoPermissionToAdd as e:
+            raise HTTPUnauthorized(content={"text": e.__repr__()})
```
masipcat, line 211: I think its better to add this here: https://github.com/plone/guillotina/blob/caa0ba163ebc1bc913f8f47aa4171b90fead81a6/guillotina/exc_resp.py, like this handler:

https://github.com/plone/guillotina/blob/caa0ba163ebc1bc913f8f47aa4171b90fead81a6/guillotina/exc_resp.py#L83-L88

nilbacardit26: okay! guillotina is too big to bear sometimes X)

masipcat (approved):

bloodbare (approved):

PR #1120, opened 2021-04-16
[5.x] No longer have dedicated read connection and do not use any prepared statements

vangheem:
The purpose of this PR is to get rid of our dedicated read connection and use of prepared statements:

- use of dedicated read conn for subset of queries can in some cases cause some lock contention between transactions
- use of prepared statements prevents people from using pgbouncer in `transaction` mode. asyncpg already automatically creates prepared statements for all sql. By removing the use of prepared statements, you can now set the statement cache size to 0 for asyncpg and have guillotina work with pgbouncer.

I also fixed memcache test issues.

### DIFF
--- a/guillotina/db/storages/pg.py
+++ b/guillotina/db/storages/pg.py
@@ -487,9 +487,6 @@ class PGConnectionManager:
     class to manage pool of connections
     """
 
-    _next_tid_sql = "SELECT nextval('{schema}.tid_sequence');"
-    _max_tid_sql = "SELECT last_value FROM {schema}.tid_sequence;"
-
     def __init__(
         self,
         dsn=None,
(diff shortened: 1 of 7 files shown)


guillotina/db/storages/pg.py
```diff
@@ -1101,11 +1069,13 @@ async def get_conflicts(self, txn):
...
                 with watch("get_conflicts_oids"):
-                    return await self.read_conn.fetch(sql, txn._tid, modified_oids)
+                    async with self.pool.acquire() as conn:
+                        return await conn.fetch(sql, txn._tid, modified_oids)
             else:
                 sql = self._sql.get("TXN_CONFLICTS", self._objects_table_name)
                 with watch("get_conflicts"):
-                    return await self.read_conn.fetch(sql, txn._tid)
+                    async with self.pool.acquire() as conn:
+                        return await conn.fetch(sql, txn._tid)
```
masipcat, line 1078: maybe it makes sense to use `conn = await txn.get_connection()` ? I guess at this point the connection is already assigned to `txn._db_conn`

vangheem: We can't. Part of the point of the read conn for these actions is it was outside the current transaction block. This operation is checking in the context outside the transaction block.

masipcat: ok, right

PR #1121, opened 2021-04-16
[master] No longer have dedicated read connection and do not use any prepared statements

vangheem:
The purpose of this PR is to get rid of our dedicated read connection and use of prepared statements:

- use of dedicated read conn for subset of queries can in some cases cause some lock contention between transactions
- use of prepared statements prevents people from using pgbouncer in `transaction` mode. asyncpg already automatically creates prepared statements for all sql. By removing the use of prepared statements, you can now set the statement cache size to 0 for asyncpg and have guillotina work with pgbouncer.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,8 @@ CHANGELOG
 6.2.7 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- No longer have dedicated read connection and do not use any prepared statements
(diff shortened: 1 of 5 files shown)


bloodbare (approved):
Nice

vangheem:
hmmm, some 3.8 tests failing. Something unrelated is wrong in tests it seems to me.

masipcat (approved):

PR #1122, opened 2021-04-19
Use connection pool when possible

vangheem:


guillotina/db/storages/pg.py
```diff
@@ -1005,16 +1028,16 @@ async def get_current_tid(self, txn):
                 return await conn.fetchval(self._max_tid_sql.format(schema=self._db_schema))
 
     async def get_one_row(self, txn, sql, *args, prepare=False, metric="get_one_row"):
-        conn = await txn.get_connection()
         # Helper function to provide easy adaptation to cockroach
         with watch(metric):
-            return await conn.fetchrow(sql, *args)
+            async with self.acquire(txn) as conn:
+                return await conn.fetchrow(sql, *args)
```
masipcat, line 1034: Looking at other methods, I see that first we acquire the connection and then we measure the query. In that case it's the other way around. I'm not sure but I think we want to acquire the connection first

vangheem: You're right! I'll fix.

masipcat (approved):

PR #1123, opened 2021-04-19
Fixing search on context on Elasticsearch

bloodbare:


masipcat (commented):
I think this is a breaking change. Maybe we should increase the minor version? So we can enforce guillotina_elasticsearch 6.1 requires guillotina 6.3

bloodbare:
> I think this is a breaking change. Maybe we should increase the minor version? So we can enforce guillotina_elasticsearch 6.1 requires guillotina 6.3

Which case would be a breaking change ? its a feature to search on a path and return all container results ?

masipcat:
> > I think this is a breaking change. Maybe we should increase the minor version? So we can enforce guillotina_elasticsearch 6.1 requires guillotina 6.3
> 
> Which case would be a breaking change ? its a feature to search on a path and return all container results ?

This change it's incompatible with the current guillotina_elasticserach that expects a container instead of a `context` (that could be a container or a folder)

masipcat (approved):

PR #1125, opened 2021-04-27
Command: run the startup/shutdown and main logic inside the loop

masipcat:
The motivation behind this change is to fix https://github.com/plone/guillotina_elasticsearch/pull/86.

### DIFF
--- a/guillotina/commands/__init__.py
+++ b/guillotina/commands/__init__.py
@@ -12,6 +12,7 @@ import asyncio
 import cProfile
 import json
 import os
+import signal
 import sys
 import yaml
 
@@ -142,7 +143,7 @@ class Command(object):
             loop_policy = resolve_dotted_name(settings["loop_policy"])
             asyncio.set_event_loop_policy(loop_policy())
 
-        app = make_app(settings=settings)
+        app = self.make_app(settings)
 
         if self.arguments.line_profiler:
             if not HAS_LINE_PROFILER:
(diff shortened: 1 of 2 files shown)


guillotina/commands/__init__.py
```diff
@@ -333,5 +309,8 @@ def command_runner():
 
     app_settings["__run_command__"] = arguments.command
     # finally, run it...
-    command = Command()
-    command.run_command()
+    try:
+        command = Command()
+        command.run_command()
+    except KeyboardInterrupt:
+        pass
```
masipcat, line 316: I think this wasn't needed in the past because of the `sys.exit(0)`

vangheem (approved):

bloodbare (changes requested):
Which is the reason to change to asyncio.run ? this does not allow to set the loop, we should change the api to not receive a loop as it can not be set on asyncio.run.

masipcat:
> Which is the reason to change to asyncio.run ? this does not allow to set the loop, we should change the api to not receive a loop as it can not be set on asyncio.run.

I didn't know we wanted to be able to change the loop and asyncio.run "just works". In fact, it takes care of some thing during the shutdown that we are not doing, and it seemed cleaner to use this function. I'm curious, which's the usecase to change the loop?

masipcat:
Now I realize that maybe the PR title it's not accurate to the intention of this PR. The important change is that now it's "asyncio-first": all the startup, main, and shutodown logic runs inside the loop.

This can be accomplished as well without using asyncio.run, and I can change it if we need to be able to change the loop

bloodbare:
> Now I realize that maybe the PR title it's not accurate to the intention of this PR. The important change is that now it's "asyncio-first": all the startup, main, and shutodown logic runs inside the loop.
> 
> This can be accomplished as well without using asyncio.run, and I can change it if we need to be able to change the loop

The use case for reusing the loop its to overwrite Command class on a class that calls multiple commands or on tests environments. In any case the contract has loop as parameter and now its not used at any place.

bloodbare:
> Now I realize that maybe the PR title it's not accurate to the intention of this PR. The important change is that now it's "asyncio-first": all the startup, main, and shutodown logic runs inside the loop.
> 
> This can be accomplished as well without using asyncio.run, and I can change it if we need to be able to change the loop

I feel ok to have the main run as async, its a change on API which will breake some code, so it should not be a bugfix release, no ?

masipcat:
> I feel ok to have the main run as async, its a change on API which will breake some code, so it should not be a bugfix release, no ?

The PR is till backwards compatible with commands with non-async run functions. But if we feel this case can be removed, I can delete the code and change the minor number

bloodbare (approved):

PR #1128, opened 2021-05-03
Fix same txn copied to all tasks/requests + Implement graceful stop (like in asyncio.run())

masipcat:
@bloodbare I've tested this locally but I'll give it a run in my test environment to be sure that it works fine. I'll ping you when this is ready for me.

### DIFF
--- a/guillotina/commands/__init__.py
+++ b/guillotina/commands/__init__.py
@@ -203,7 +203,6 @@ class Command(object):
             try:
                 self._cancel_all_tasks(self.loop)
                 self.loop.run_until_complete(self.loop.shutdown_asyncgens())
-                self.loop.run_until_complete(self.loop.shutdown_default_executor())
             finally:
                 self.loop.close()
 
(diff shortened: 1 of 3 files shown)


guillotina/commands/__init__.py
```diff
@@ -191,17 +190,26 @@ def __run_with_monitor(self, app, settings):
...
+                self._cancel_all_tasks(self.loop)
+                self.loop.run_until_complete(self.loop.shutdown_asyncgens())
+            finally:
+                self.loop.close()
 
     async def __run_async(self, app, settings):
-        await app.startup()
+        # We run app.startup() in another task to prevent assigning values to the contextvars in the 'main task'
+        # Without this change the 'txn' (and all other ctxvars) are copied and shared to all requests
+        await asyncio.create_task(app.startup())
```
vangheem, line 212: oh, nice

vangheem (approved):
I wonder if we could have some unit tests for this?

masipcat:
> I wonder if we could have some unit tests for this?

I'll write one later today

bloodbare (approved):

PR #1132, opened 2021-05-14
PG catalog unrestriced search and parser

rboixaderg:
Added unrestriced in build_count_query function

Modify pg catalog parser to use __or operator. 

I don't see any place that explain how we can use __or operator, then I modify it to allow recieve an encoded query

### DIFF
--- a/guillotina/contrib/catalog/pg/parser.py
+++ b/guillotina/contrib/catalog/pg/parser.py
@@ -34,7 +34,7 @@ class Parser(BaseParser):
         wheres = []
         arguments = []
         selects = []
-        for _, (sfield, svalue) in enumerate(parsed_value):
+        for sfield, svalue in parsed_value:
             result = self.process_queried_field(sfield, svalue)
             if result is not None:
                 wheres.append(result[0])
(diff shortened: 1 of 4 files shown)


guillotina/contrib/catalog/pg/parser.py
```diff
@@ -27,19 +28,20 @@ class ParsedQueryInfo(BasicParsedQueryInfo):
...
     def process_compound_field(self, field, value, operator):
-        if not isinstance(value, dict):
+        parsed_value = urllib.parse.parse_qsl(urllib.parse.unquote(value))
+        if not isinstance(parsed_value, list):
             return None
         wheres = []
         arguments = []
         selects = []
-        for sfield, svalue in value.items():
+        for _, (sfield, svalue) in enumerate(parsed_value):
```
masipcat, line 37:
```suggestion
        for sfield, svalue in parsed_value:
```

rboixaderg: changed!

bloodbare (approved):

PR #1146, opened 2021-07-19
Fix vocabulary country code AN

jotare:


bloodbare:
AFAIK the code for it is ANT and for Andorra AD. If you want to update the PR we will merge

bloodbare (changes requested):
ANT for Netherlands Antilles and AD for Andorra

jotare:
> AFAIK the code for it is ANT and for Andorra AD. If you want to update the PR we will merge

Yes, but that's the Alpha-3 code for this country. It's a good idea to change a country to a 3 character code? The official ISO website says that the Alpha-2 code for Netherlands Antilles is AN: 
https://www.iso.org/obp/ui/#iso:code:3166:AN

PR #1154, opened 2021-10-28
Use connection pool when possible (#1122)

masipcat:
Ported https://github.com/plone/guillotina/pull/1122


guillotina/db/storages/pg.py
```diff
@@ -1135,77 +1148,67 @@ async def get_annotation_keys(self, txn, oid):
...
+        sql = self._sql.get("NUM_RESOURCES_BY_TYPE", self._objects_table_name)
+        async with self.acquire(txn) as conn:
             with watch("total_objects_by_type"):
                 result = await conn.fetchval(sql, type_)
         return result
 
     # Massive treatment without security
     async def _get_page_resources_of_type(self, txn, type_, page, page_size):
-        conn = await txn.get_connection()
-        async with txn._lock:
```
masipcat, line 1208: @masipcat recuperar

guillotina/db/strategies/resolve.py
```diff
@@ -1,56 +0,0 @@
...
-    which means we have shorter db transactions but reads might not be inconsistent.
-    """
-
-    async def tpc_begin(self):
-        pass
-
-    async def tpc_commit(self):
-        await self.retrieve_tid()
-        if self._transaction._db_txn is None:
-            await self._storage.start_transaction(self._transaction)
```
masipcat, line 56: @bloodbare I was reviewing this part and I don't understand why the strategy `resolve_readcommitted` (the default one) needs the `tpc_vote` logic from `resolve` strategy.

If I'm not wrong, these are the cases we are trying to mitigate:

1. Cache invalidation race condition: an object is modified but the next request that wants to modify the object reads the stale object. During the write to the storage, we want to detect that the old txn id is not the latest in postgres. This is already implemented in the UPDATE sql, using the `WHERE tid = $7::bigint`. If the WHERE clause doesn't match it would not update the object and we'll raise the ConflictError.

2. Two concurrent requests modify the same object "at the same time". At the same time could mean two things:
a) Both requests read the same object, the first starts the pg transaction and writes the changes and finished. When the second transaction starts will try to update the object and I'd would happen the error in point 1. The Old TID doesn't match and the ConflictError is raised.
b) Both requets read the same object, the first request starts the transaction and writes the changes. At the same time, before committing the pg transaction the second request starts a transaction and tries to write the same object. In that case two things could happen: 1. same thing as previous requests. 2. deadlock and both requests fail

My conclusion is that in all these situations postgres/SQL queries already handle the conflicts, and there is no need for the extra logic. The only case I could think this is useful is for Cockroach DB.

./cc @vangheem

masipcat: Since this is something that didn't change in this PR, I move this discussion to an issue to review it in the future

bloodbare (approved):

PR #1160, opened 2021-12-01
Breaking change: guillotina futures won't have the txn set anymore

masipcat:
Before this PR the txn used in the request was reused in all the futures and in some cases this produced race conditions and unpredictable behavior.


contrib-requirements.txt
```diff
@@ -8,6 +8,7 @@ mypy-zope==0.3.2
 black==21.10b0
 isort==4.3.21
 jinja2==2.11.3
+MarkupSafe<2.1.0
```
masipcat, line 11: version 2.1 removed 'soft_unicode' and jinja2 depends on it

bloodbare (approved):

PR #1161, opened 2021-12-02
Improve error logging in futures

masipcat:
@bloodbare I created a new notice_on_error function. I wasn't sure if it's ok to change the behavior of the existing function

### DIFF
--- a/guillotina/contrib/cache/strategy.py
+++ b/guillotina/contrib/cache/strategy.py
@@ -8,7 +8,7 @@ from guillotina.exceptions import NoChannelConfigured
 from guillotina.exceptions import NoPubSubUtility
 from guillotina.interfaces import ICacheUtility
 from guillotina.profile import profilable
-from guillotina.utils import notice_on_error2
+from guillotina.utils import notice_on_error_internal
 from typing import Any
 from typing import Dict
 from typing import List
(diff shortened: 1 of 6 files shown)


guillotina/utils/misc.py
```diff
@@ -409,12 +409,16 @@ def get_request_scheme(req) -> str:
 
 
 async def notice_on_error(key: str, func_to_await):
+    return await notice_on_error2(f"Error on initialize utility {key}", func_to_await)
+
+
+async def notice_on_error2(msg: str, func_to_await):
```
bloodbare, line 415: A better naming, `notice_on_error_internal`

guillotina/contrib/cache/strategy.py
```diff
@@ -129,7 +130,9 @@ async def close(self, invalidate=True, publish=True):
                     await self.fill_cache()
                     if len(self._keys_to_publish) > 0 and self._utility._subscriber is not None:
                         keys = self._keys_to_publish
-                        asyncio.ensure_future(self.synchronize(keys))
+                        asyncio.ensure_future(
+                            notice_on_error2("Error on cache synchronize", self.synchronize(keys))
```
bloodbare, line 134: Why not `notice_on_error`

guillotina/utils/misc.py
masipcat: ok!

guillotina/contrib/cache/strategy.py
masipcat: Because it would log `Error on initialize utility {...}`. I guess this message it only makes sense when calling `utility.initialize()`

bloodbare (approved):

PR #1169, opened 2022-03-29
Fix PubSub utility stuck in infinite loop when Redis connection is interrupted

masipcat:


guillotina/contrib/pubsub/utility.py
```diff
@@ -58,6 +58,12 @@ async def finalize(self, app):
     async def real_subscribe(self, channel, channel_name):
         while channel_name in self._subscribers:
             try:
+                if channel is None:
+                    # wait 1 second and try to connect
+                    await asyncio.sleep(1)
```
vangheem, line 63: Hhhm. This means I'm startup now there will always be a 1 second delay before this starts.

vangheem: Should the sheep still be in exception handler but we just remove the finally block like you did?

vangheem (approved):
My only comment is that now startup is delayed for that

guillotina/contrib/pubsub/utility.py
masipcat: > Hhhm. This means I'm startup now there will always be a 1 second delay before this starts.

No, the channel is initialized here:

https://github.com/plone/guillotina/blob/98ee68b1c3c1a09bc5f65f9fc43d4e370a55843d/guillotina/contrib/pubsub/utility.py#L96-L97

masipcat: > Should the sheep still be in exception handler but we just remove the finally block like you did?

I'm not 100% sure but the problem was that after catching the ConnectionError, the next time wasn't crashing, just iterating the empty async generator thus not executing the sleep inside the exception and producing the cpu spinning.

vangheem: Ahh, I gotcha. That's a bit confusing.

PR #1171, opened 2022-05-30
Deserialize pickles in a asyncio executor + converted db functions reader() and IWriter.serialize() to async

masipcat:


guillotina/db/writer.py
```diff
@@ -56,7 +56,7 @@ def old_serial(self):
     def part(self):
         return getattr(self._obj, "__partition_id__", 0)
 
-    def serialize(self):
+    async def serialize(self):
```
vangheem, line 59: Does this need to change?

vangheem (approved):
I've always thought about this.

Making the reader interface async will also allow for more flexibility if you need to hook into anything else during object reading.

async is like an infection though -- the number of interfaces this affects is a shame...

IMO, this is a breaking change. Lots of interfaces that are potentially used.

We could make the interface optionally async. Use https://github.com/plone/guillotina/blob/master/guillotina/utils/misc.py#L126 to not cause breaking change

masipcat:
> IMO, this is a breaking change. Lots of interfaces that are potentially used.

OK, agree

> We could make the interface optionally async. Use https://github.com/plone/guillotina/blob/master/guillotina/utils/misc.py#L126 to not cause breaking change

I'm ok with this change but third-party applications using the guillotina IWriter will still break so maybe we should update the minor version?

guillotina/db/writer.py
masipcat: I didn't add the same logic as in the deserialization function (because it's expensive to know if a python object it's big or small) but I changed the function to async just in case someone wants to use a custom logic to decide when to run_async. For example:
 
```python
async def serialize(obj):
    if isinstance(obj, BigObject):
        return await run_async(pickle.dumps, obj)
    else:
        return pickle.dumps(obj)
```

masipcat: But if we feel it's not necessary I can revert this part

PR #1175, opened 2022-08-24
being able to call do_action with super powers using the utility

nilbacardit26:
In some situations in my internal code, I want to be able to switch the state of the workflow on behalf of other users without checking any security policies.

### DIFF
--- a/guillotina/contrib/workflows/utility.py
+++ b/guillotina/contrib/workflows/utility.py
@@ -60,17 +60,18 @@ def create_workflow_factory(proto_name, proto_definition):
         def initial_state(self):
             return self._initial_state
 
-        async def do_action(self, action, comments):
+        async def do_action(self, action, comments, bypass_permission_check=False):
             available_actions = self.actions
             if action not in available_actions:
                 raise HTTPPreconditionFailed(content={"reason": "Unavailable action"})
 
             action_def = available_actions[action]
-            policy = get_security_policy()
-            if "check_permission" in action_def and not policy.check_permission(
-                action_def["check_permission"], self.context
-            ):
-                raise HTTPUnauthorized()
+            if bypass_permission_check is False:
+                policy = get_security_policy()
+                if "check_permission" in action_def and not policy.check_permission(
+                    action_def["check_permission"], self.context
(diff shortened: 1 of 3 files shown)


guillotina/contrib/workflows/utility.py
```diff
@@ -60,17 +60,18 @@ async def available_actions(self, request):
         def initial_state(self):
             return self._initial_state
 
-        async def do_action(self, action, comments):
+        async def do_action(self, action, comments, root=False):
```
vangheem, line 63: What do you think about renaming `root` to `bypass_permission_check` so it's more explicit?

vangheem (changes requested):

guillotina/contrib/workflows/utility.py
nilbacardit26: sure @vangheem much better

vangheem (changes requested):
Would you be able to run the file through black? It's failing the formatting check.

nilbacardit26:
yeap, now it should be fine

nilbacardit26:
seems that there are some tests failing including the ones of the workflow, let me check this out @vangheem

nilbacardit26:
seems like there is an error with the pytest_docker_fixtures, related to docker. For some reason, when trying to pull/run the docker we get a 404:
`404 Client Error: Not Found for url: http+docker://localhost/v1.23/containers/create` 

I need to dig a little bit more, it works fine on my computer. It might be related to that:
https://github.com/docker/docker-py/issues/2101#issuecomment-939976795

vangheem (approved):

vangheem:
oh dang, CI is busted!

I won't have time to look at it to fix unfortunately.

@nilbacardit26 will you be able to use a manual release from your branch while it gets sorted?

nilbacardit26:
@vangheem yes, that would suffice! Thank you

masipcat:
I think pinning `docker` (dockerpy) to < 6.0 will fix the issue

nilbacardit26:
@masipcat thank you! now it is working :)

PR #1179, opened 2022-10-20
adding slash in path__starts queries when context is not the container

nilbacardit26:
I found bug when searching with pg catalog on contexts that starts with the same path than others:
```
/db/container/foo_folder/Item1
/db/container/foo_folder2/Item2

/db/container/foo_folder/@search
```
it returned Item1 and Item2

### DIFF
--- a/guillotina/catalog/parser.py
+++ b/guillotina/catalog/parser.py
@@ -83,9 +83,11 @@ class BaseParser:
         # Path specific use case
         if "path__starts" in params:
             path = params.pop("path__starts")
-            path = "/" + path.strip("/")
+            path = "/" + path.lstrip("/")
         else:
             path = get_content_path(self.context)
+            if path.endswith("/") is False and path != "/":
+                path = path + "/"
         params["path__starts"] = path
 
         if "_size" in params:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,10 @@ CHANGELOG
 - Deps: replace aioredis with redis-py
 - Deps: updated flake8 so it won't depend on 'importlib-metadata<5'
   [masipcat]
+- Fix path__starts. Add a slash when parsing the path of the query if
+  the context of the search is not the container, to avoid getting the
+  results of contexts that starts with the same path.
+  [nbacardit26]
 
 
 6.4.2 (2022-08-25)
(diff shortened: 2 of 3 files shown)


guillotina/catalog/parser.py
```diff
@@ -86,6 +86,8 @@ def __call__(self, params: typing.Dict) -> BasicParsedQueryInfo:
             path = "/" + path.strip("/")
         else:
             path = get_content_path(self.context)
+        if path.endswith("/") is False and path != "/":
+            path = path + "/"
```
masipcat, line 90: I'm not 100% sure if we want to add this trailing slash in all cases or only when the path is defined implicitly based on context.

I could see someone specifying `path__starts` explicitly without the trailing slash and expect the current behavior (items from different folders that start with same prefix)

nilbacardit26: yes! I agree @masipcat this could be a good functionality to have, my thought though when I did this, was to preserve the same behavior when not passing `paths__starts`, which in some cases can lead to confusions on how it works. What is clear is that by calling `/db/container/foo_folder/@search` I am not expecting to get any results from another folder that is not `foo_folder`. 
Do you think that is worthy to have this dual behavior when passing `path__starts`? If we all agree, we just have to specify it somewhere, or it can lead to miss behaviors.

masipcat: I wouldn't say it's a dual behavior. We are not changing how the filter `path__starts` behaves. We are changing how the implicit search in a folder works. And as you said, it's clear that the current behavior of `@search` in a folder doesn't work as expected.

But I'd like to know what @bloodbare and @vangheem think about this as well

masipcat: Also it's funny the code in L86 because right now the trailing slash is always stripped. I'm not sure if this behavior is intended or should've been `path = "/" + path.lstrip("/")`

nilbacardit26: You are right @masipcat . If you all agree and want to have the same behavior as now, but just adding the slash when using the context, it is fine for me, however, I think that the trailing slash being removed when injecting path__starts on the query for me is not the behavior the user expects. I think that what you propose with the line `path = "/" + path.lstrip("/")` is the way to go. We let the user choose how he wants to search.

nilbacardit26: @masipcat the last commit does what we have talked X)

masipcat (approved):

bloodbare (approved):

PR #1186, opened 2023-10-10
adding OrderedDict Field

nilbacardit26:


bloodbare (changes requested):
I would change the API endpoint to something more meaninfull. From `@orderImages` to `@sort`

PR #1189, opened 2023-11-16
Deleting import zope.interface implements

nilbacardit26:
Every time I install guillotina as a dependency I encounter this error:
Error importing plugin "guillotina.tests.fixtures": cannot import name 'implements' from 'zope.interface. implements is not there anymore in version 6.1 'zope.interface

I do not know what this line does, implements is not used anywhere. I passed all the test without this line. Some of you know why is it there?
https://github.com/plone/guillotina/blob/master/guillotina/component/interfaces.py#L19

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,8 @@ CHANGELOG
 6.4.4 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Removing implements import in components/interfaces.py
+  [nilbacardit26]
 
 
 6.4.3 (2023-10-11)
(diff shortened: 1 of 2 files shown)


vangheem:
It's a vanity import to be able to use the `implements` interface. If that is no longer used/support, remove the import!

I don't think it's a good idea to hard pin a version -- it can be annoying for consumers

nilbacardit26:
@vangheem done, I will run the tests I have in another project just to be sure this import is not breaking anything

vangheem (approved):

masipcat (approved):

vangheem:
@nilbacardit26 everything look good?

We can merge then

nilbacardit26:
@vangheem yes, I am not authorized to merge, feel free to do it

PR #1192, opened 2023-12-02
Fix dbusers - Groups and indexed data

rboixaderg:


masipcat (commented):
Looks good to me, but I think this PR should include a migration to reindex the JSON field

rboixaderg:
I need to review this part, because maybe we have an error in catalog when we use index_name property, also this affect in guillotina_elasticsearch.

PR #1208, opened 2026-01-30
add async context to acquire advisory locks

nilbacardit26:
We need to lock an object to prevent concurrent modifications by other transactions. I've added this piece of code to use pessimistic advisory locking, using an async context manager.


bloodbare (commented):
As its not a core DB functionallity used on guillotina, maybe can be an external utility. G is designed to not need the locks.

PR #1211, opened 2026-02-27
Add low-level MCP contrib package with tools and tests

nilbacardit26:
WIP


guillotina/contrib/mcp/subscribers.py
```diff
@@ -0,0 +1,18 @@
+from guillotina import configure
+from guillotina.component import query_utility
+from guillotina.contrib.mcp.interfaces import IMCPToolRegistry
+from guillotina.interfaces import IBeforeObjectRemovedEvent
+from guillotina.interfaces import IObjectAddedEvent
+from guillotina.interfaces import IObjectModifiedEvent
+from guillotina.interfaces import IResource
+
+
+@configure.subscriber(for_=(IResource, IObjectAddedEvent))
```
bloodbare, line 12: Does not work on distributed systems

nilbacardit26: @bloodbare take another look now, please. I changed that to use redis if available with an asyncio queue to avoid spamming the invalidation of the cache too much

nilbacardit26:
@bloodbare fear feel not for the 300+ files changed. MCP has a higher dependency on uvicorn than the one we already have, so I decided to upgrade uvicorn to the latest version (0.42.0) and drop support for 3.8 and 3.9 python versions, as they are too old to be maintained with the new uvicorn and the whole stack. I added test coverage for version 3.12. Tests are passing.

I upgraded the versions of the test packages and lint tools as well: isort and black. Most of the files are due to black and isort changes. Async fixtures decorators are changed to the most modern syntax as well.

Adding MCP has been a good reason I believe to do all these upgrades

Maybe now that I am doing this, I can add support to at least 3.13 too.

bloodbare (approved):

